PR #9, opened 2021-09-01
Improve documentation for docker-compose and initial setup

kaladinlight:


phewitt (approved):
I think this makes it more obvious what is required.

PR #34, opened 2021-09-17
Update API to use /account/{pubkey} pathing

### LINKED ISSUE #35: Unchained API: More restful design
Instead of lots of flat endpoints like `/balance`, `/utxos`, etc., use a more resource driven restful design pattern like:

`/account/{pubkey}`
`/account/{pubkey}/txs`
`/account/{pubkey}/utxos`
`/gas/estimate`
`/gas/price`

kaladinlight:


coinstacks/ethereum/api/src/controller.ts
```diff
@@ -72,8 +74,9 @@ export class Ethereum extends Controller implements BaseAPI, EthereumAPI {
       }, [])
 
       return {
-        pubkey: data.address,
         balance: data.balance,
+        nonce: Number(data.nonce ?? 0),
```
phewitt, line 78: is it always safe to assume undefined nonce means start at 0?

docker-compose.yml
```diff
@@ -8,14 +8,13 @@ services:
       BROKER_URL: amqp://guest:guest@rabbitmq:5672
       MONGO_URL: mongodb://mongo:27017
     working_dir: /app/coinstacks/${COINSTACK}/api
-    command: dockerize -wait tcp://mongo:27017 -timeout 5m -wait-retry-interval 10s sh -c "yarn nodemon"
+    command: yarn nodemon
```
phewitt, line 11: not waiting for mongo anymore?

phewitt (approved):
Idk how much I like committing generated code. It really makes a lot of noise on a PR

docker-compose.yml
kaladinlight: Removed the register/unregister endpoints at this time since they aren't being used. Will most likely be adding back in some way come websocket subscription

mvpratt: still needed for block height though yeah?

coinstacks/ethereum/api/src/controller.ts
kaladinlight: Think it is more of just a typing things since `nonce` is optional as in only eth uses it for blockbook. Every address has a nonce, so my assumption was it should technically always return, but if it doesn't I am just assuming new address and should start at 0 in that case.

kaladinlight:
> Idk how much I like committing generated code. It really makes a lot of noise on a PR

Yep, this has been mentioned and I agree. Made this issue to address: https://github.com/shapeshift/unchained/issues/36

coinstacks/bitcoin/sample.env
```diff
@@ -0,0 +1,11 @@
+# SECRET ENVIRONMENT VARIABLES
```
mvpratt, line 1: 👍

docker-compose.yml
kaladinlight: In the ingester, this is just the dockerize wait for the API service

packages/client/src/generated/ethereum/api.ts
```diff
@@ -276,7 +282,7 @@ export const V1ApiAxiosParamCreator = function (configuration?: Configuration) {
             assertParamExists('estimateGas', 'to', to)
             // verify required parameter 'value' is not null or undefined
             assertParamExists('estimateGas', 'value', value)
-            const localVarPath = `/api/v1/estimate-gas`;
+            const localVarPath = `/api/v1/gas/estimate`;
             // use dummy base URL string because the URL constructor only accepts absolute URLs.
```
mvpratt, line 286: would it make more sense to do gas estimation in the client?

packages/client/src/generated/ethereum/api.ts
```diff
@@ -312,15 +318,15 @@ export const V1ApiAxiosParamCreator = function (configuration?: Configuration) {
             };
```
mvpratt, line 318: do we have an endpoint to get network fees?

mvpratt (approved):

packages/client/src/generated/ethereum/api.ts
kaladinlight: yes, we have `/gas/price` endpoint: https://github.com/shapeshift/unchained/blob/244f05ff15bfd0ebf35a14a83b702a36cf9c7d88/coinstacks/ethereum/api/src/controller.ts#L192

packages/client/src/generated/ethereum/api.ts
kaladinlight: :shrug: either or would work. was already added for chain adapters to use so just carrying it forward

PR #49, opened 2021-09-27
Address index

### LINKED ISSUE #64: Add the necessary endpoints that will be needed by chain-adapters
We need a `getTransaction` endpoint to grab transaction data, and whatever else that is needed by chain-adapters to send Bitcoin transactions.

kaladinlight:


coinstacks/common/api/src/models.ts
```diff
@@ -33,6 +33,7 @@ export interface InternalServerError {
 export interface Account {
   balance: string
   pubkey: string
+  path?: string
```
kaladinlight, line 36: Add path to `BitcoinAccount`

packages/blockbook/src/models.ts
```diff
@@ -257,8 +257,8 @@ export interface Utxo {
   value: string
   height?: number
   confirmations: number
-  address?: string
-  path?: string
+  address: string
```
kaladinlight, line 260: These types are modeled directly off the blockbook types and should stay optional

mvpratt (approved):

packages/blockbook/src/controller.ts
```diff
@@ -642,15 +642,6 @@ export class Blockbook extends Controller {
    * @example account "14mMwtZCGiAtyr8KnnAZYyHmZ9Zvj71h4t"
    * @example account "xpub6DQYbVJSVvJPzpYenir7zVSf2WPZRu69LxZuMezzAKuT6biPcug6Vw1zMk4knPBeNKvioutc4EGpPQ8cZiWtjcXYvJ6wPiwcGmCkihA9Jy3"
    */
-  @Example<Array<Utxo>>([
```
kaladinlight, line 645: And this is still a valid response for utxo by address

packages/blockbook/src/controller.ts
```diff
@@ -277,9 +277,9 @@ export class Blockbook extends Controller {
     },
   })
   @Get('tx-specific/{txid}')
-  async getTransactionSpecific(@Path() txid: string): Promise<TxSpecific> {
+  async getTransactionSpecific(@Path() txid: string): Promise<Tx> {
     try {
-      const { data } = await this.instance.get<TxSpecific>(`api/v2/tx-specific/${txid}`)
+      const { data } = await this.instance.get<Tx>(`api/v2/tx-specific/${txid}`)
```
kaladinlight, line 282: These should remain `TxSpecific` and we will need to add each chain's specific interface. In this case we can add `BitcoinTxSpecific`

packages/blockbook/src/controller.ts
```diff
@@ -229,7 +229,7 @@ export class Blockbook extends Controller {
    *
    * @example txid "0xf5ef950d597bea8c980fe5a17faeb70ed7f5030dcf58f077baede5eb0d610cdf"
    */
-  @Example<TxSpecific>({
+  @Example<EthTxSpecific>({
```
kaladinlight, line 232: This is good, we should be showing the corresponding ethereum specific type for the ethereum example. We can add each additional example as coins are added.

```
2. Bitcoin

@example txid "bitcointxid"

@Example<BitcoinTxSpecific>
```
In the appropriate ordering with eth

coinstacks/common/api/src/models.ts
kaladinlight: Still want to pull this out and extend `BitcoinAccount`

coinstacks/bitcoin/api/src/controller.ts
```diff
@@ -58,23 +58,44 @@ export class Bitcoin extends Controller implements BaseAPI, BitcoinAPI {
     try {
       let data: Address | Xpub
       if (isXpub(pubkey)) {
-        data = await blockbook.getXpub(pubkey, undefined, undefined, undefined, undefined, 'tokenBalances', 'used')
+        data = await blockbook.getXpub(pubkey, undefined, undefined, undefined, undefined, 'txs', 'used')
```
kaladinlight, line 61: No need to fetch verbose transactions anymore
```suggestion
        data = await blockbook.getXpub(pubkey, undefined, undefined, undefined, undefined, 'tokenBalances', 'used')
```

packages/blockbook/src/controller.ts
SterL: yep all UtxoResponse is pulled out

packages/blockbook/src/controller.ts
SterL: yep all UtxoResponse is removed now

coinstacks/bitcoin/api/src/models.ts
```diff
@@ -78,14 +78,14 @@ export type BtcTxSpecific = {
 /**
  * Contains info about an unspent transaction output
  */
-export interface UtxoResponse {
+export interface Utxo {
```
kaladinlight, line 81: We can update this type to be our own shape that we want unchained to respond with. Where I have been starting with this strategy is only passing back what we are actually using in the chain adapter side, and then maybe making the `pubkey` vernacular abstraction if we want instead of separating `address` and `path`. This will allow us to leverage blockbook, but build out our own API how we see fit and try not to just copy blockbook because we happen to be using blockbook

coinstacks/common/api/src/models.ts
SterL: Ok my bad I see what you're sayin.

kaladinlight: Looks like it was added, so we can just remove

coinstacks/bitcoin/api/src/controller.ts
```diff
@@ -169,6 +190,7 @@ export class Bitcoin extends Controller implements BaseAPI, BitcoinAPI {
       txid: '02cdb69a97d1b8585797ac31a1954804b40a71c380a3ede0793f21a2cdfd300a',
       value: '729',
       vout: 1,
+      path: "m/44'/0'/0'/0/2",
```
kaladinlight, line 193: This should ideally be 2 example responses, one for the address response (no path or address) and one for the xpub response (includes address and path)

coinstacks/bitcoin/api/src/controller.ts
```diff
@@ -178,16 +200,85 @@ export class Bitcoin extends Controller implements BaseAPI, BitcoinAPI {
...
+      confirmations: 498,
+      time: 1632513682,
+      blocktime: 1632513682,
+    },
+  ])
+  @Response<BadRequestError>(400, 'Bad Request')
+  @Response<ValidationError>(422, 'Validation Error')
+  @Response<InternalServerError>(500, 'Internal Server Error')
+  @Get('transaction/{txid}')
+  async getTransaction(@Path() txid: string): Promise<BtcTxSpecific> {
```
kaladinlight, line 278: The one case where I don't know about trying to trim down the response would be the transaction payload from the node directly. Since we should be typing this out in the blockbook package, we can import from there.

coinstacks/bitcoin/api/src/models.ts
```diff
@@ -1,25 +1,123 @@
...
+export interface BitcoinAccount extends Account {
+  /**
+   * Account details by address if BitcoinAccount was fetched by xpub
+   */
+  addresses?: Array<Account>
+
+  /**
+   * Transactions of the given xpub
+   */
+  transactions?: Array<Tx>
```
kaladinlight, line 105: This can be removed with no longer requesting `txs` details from getAccount

packages/blockbook/src/controller.ts
SterL: Roger that. Meant to include this.

packages/blockbook/src/controller.ts
```diff
@@ -229,7 +229,7 @@ export class Blockbook extends Controller {
    *
    * @example txid "0xf5ef950d597bea8c980fe5a17faeb70ed7f5030dcf58f077baede5eb0d610cdf"
    */
-  @Example<TxSpecific>({
```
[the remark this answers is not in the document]
SterL: Yep So these are just examples of what the responses looks like. I forgot to include the Bitcoin example response. It doesn't effect the code afaik. But it does generate the swagger docs/ui, and makes things convenient there.

coinstacks/bitcoin/api/src/controller.ts
natven: does that mean returning a blockbook type in our api? IMO our api should be 100% decoupled from backends it depends on, even if that means creating a type that is essentially a duplicate of the upstream type

coinstacks/bitcoin/api/src/models.ts
natven: are you asking for changes here? the attributes I see there make sense for a utxo at a glance. even though chain-adapters doesn't use `coinbase` for example that's a standard enough thing to include up front IMO

coinstacks/bitcoin/api/src/models.ts
natven: mostly agree I think. is it worth including txids? concern that would cause a burden for other chain implementations?

coinstacks/bitcoin/api/src/models.ts
kaladinlight: With all of the other types, I have been starting slim with what we as currently the only client are using and more can easily be added if there is need for the information by external clients. We can include all of it, but the point I am trying to drive in as we build out this API is that we should be thinking about our own unchained abstraction and not using blockbook types verbatim just because they exist and are the easy crutch to use. I want to separate our API abstraction from blockbook's abstraction.

kaladinlight: Utxo is pretty generic so they might end up being very similar, I just want us to be cognizant of the intention of trying to build out our own types for our unchained API abstraction.

packages/blockbook/src/models.ts
```diff
@@ -135,34 +135,6 @@ export interface EthereumSpecific {
   data?: string
 }
 
-/**
- * Contains ethereum specific transaction info as returned from the node
- */
-export interface EthTxSpecific {
```
phewitt, line 141: :+1: removing chain specifics from blockbook

phewitt (approved):
lets get it in!

PR #50, opened 2021-09-27
AWS Node Termination Handler

### LINKED ISSUE #41: Use k8s spot termination handler
https://github.com/pusher/k8s-spot-termination-handler

Currently, spot inst termination signals are not caught and handled resulting in corrupted dbs during sync if nodes churn.

AC:
- Handle spot termination signal and gracefully shutdown pods

samortaylor:
Update cluster-launcher to leverage AWS node termination handler.  This is to prevent SPOT instances from being preempted without the ability to gracefully stop pods running on affected nodes. 

https://github.com/shapeshift/cluster-launcher/pull/8

### DIFF
--- a/pulumi/package.json
+++ b/pulumi/package.json
@@ -7,7 +7,7 @@
     "clean": "rm -rf dist node_modules"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


phewitt (approved):
lgtm

PR #75, opened 2021-10-04
use root address if there is no nextAddressIndex

SterL:
If an xpub does not have any tx history, it will not have a `tokens` object in the response of `getXPub`, which means that we need to use 0 for the `nextReceiveAddressIndex` rather than undefined.

### DIFF
--- a/coinstacks/bitcoin/api/src/controller.ts
+++ b/coinstacks/bitcoin/api/src/controller.ts
@@ -83,8 +83,8 @@ export class Bitcoin extends Controller implements BaseAPI, BitcoinAPI {
       return {
         pubkey: data.address,
         balance: data.balance,
-        nextReceiveAddressIndex: nextAddressIndexes[0] || 0,
-        nextChangeAddressIndex: nextAddressIndexes[1] || 0,
+        nextReceiveAddressIndex: nextAddressIndexes[0] ?? 0,
+        nextChangeAddressIndex: nextAddressIndexes[1] ?? 0,
       }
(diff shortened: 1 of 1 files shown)


coinstacks/bitcoin/api/src/controller.ts
```diff
@@ -83,8 +83,8 @@ export class Bitcoin extends Controller implements BaseAPI, BitcoinAPI {
       return {
         pubkey: data.address,
         balance: data.balance,
-        nextReceiveAddressIndex: nextAddressIndexes[0],
-        nextChangeAddressIndex: nextAddressIndexes[1],
+        nextReceiveAddressIndex: nextAddressIndexes[0] || 0,
+        nextChangeAddressIndex: nextAddressIndexes[1] || 0,
```
kaladinlight, line 87: prefer `??` over `||` for default assignment on nullish value
```suggestion
        nextReceiveAddressIndex: nextAddressIndexes[0] ?? 0,
        nextChangeAddressIndex: nextAddressIndexes[1] ?? 0,
```

kaladinlight: If we are using a default, we may want to construct the return payload checking the `isXpub` value since we will now also be returning address indexes when fetching account details by address which isn't accurate.

kaladinlight: The other alternative is the client can chose their starting index if undefined

cjthompson (approved):

PR #87, opened 2021-10-11
Add required from query parameter to estimateGas

### LINKED ISSUE #84: estimateGas: support from address
From address is an optional argument that can be sent to the `estimateGas` call and may have implications on the estimated gas cost returned.

AC:
- Support a `from` query paramter for `estimateGas` endpoint

kaladinlight:


0xdef1cafe (approved):
🙏

PR #96, opened 2021-10-14
Add BTC fee estimation endpoint

### LINKED ISSUE #63: Unchained: Add BTC fee estimation
The BTC chain adapter currently fetches network fee estimations from bitcoinfees.earn.com. This is effective, but fragile in the case of unforeseen external service interruptions. Unchained should provide BTC fee estimates via API call, with the network fees determined independently, without the use of any external service. For reference, look at how BTC fees are estimated in Watchtower.

pastaghost:


kaladinlight (approved):
Looks great, thank you! :100:

PR #110, opened 2021-10-21
use traefik to load balance multiple coinstacks

phewitt:
It has become apparent that we need to be able to run multiple coinstacks locally. At first this would have required exposing each service at different ports.

So I decided to use a reverse proxy to handle all traffic to all container. Now the only ports exposed are port `80` for the reverse proxy and port `8080` for the proxies dashboard. all routing to coinstacks can be done via localhost now. For example

`api.bitcoin.localhost/docs`
`rabbit-admin.ethereum.localhost`
`mongo.bitcoin.localhost`

This should simplify networking locally.

To use this now simply run `docker-compose up` in the root of the project. Then run `docker-compose up` in the coinstacks you want to run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,6 @@ Traefik routes requests based on host name. which includes the coinstack name. F
 #### **Prerequisites**
 
 - Install [docker-compose](https://docs.docker.com/compose/install/)
-- To start up the reverse proxy and hot reloading for files run from the root of the project
-  ```sh
-  docker-compose up
-  ```
 - Copy sample env file:
   ```sh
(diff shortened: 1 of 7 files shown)


kaladinlight (approved):

mvpratt (approved):

elmutt (approved):

elmutt (changes requested):
Need to remove `unchained-local` from the `depends_on:` sections of bitcoin and ethereum `docker-compose.yml` for this to work.

Otherwise it gives:

```ERROR: Service 'api' depends on service 'unchained-local' which is undefined.```

elmutt (approved):

PR #121, opened 2021-10-27
use RxObservable to encapsulate subscription

mrnerdhair:
Not yet tested, but should demonstrate how much easier Observables make handling subscriptions to data feeds.


mrnerdhair:
stale

PR #126, opened 2021-10-29
Clustered RabbitMQ

### LINKED ISSUE #111: RabbitMQ Cluster/Replicas
There is currently only a single rabbitmq sts deployed for each coinstack which will result in down time whenever kube nodes roll. It would be best if we could update to use a rabbitmq cluster or replica set deployment, similar to mongodb

phewitt:
Adds a RabbitMQ Cluster Operator custom resource definition so that our coinstacks can leverage a highly available RabbitMQ cluster.


coinstacks/bitcoin/docker-compose.yml
```diff
@@ -1,5 +1,9 @@
 version: '3.6'
 
+x-env: &env
```
kaladinlight, line 3: Cool!

coinstacks/bitcoin/pulumi/Pulumi.sample.yaml
```diff
@@ -33,6 +33,13 @@ config:
       replicaCount: 1
       storageSize: '1Gi'
 
+    ## rabbit - if specified a rabbit cluster will be deployed
+    rabbit:
+      cpuLimit: '500m'
+      memoryLimit: '512Mi'
+      replicaCount: 1
+      storageSize: '5Gi'
```
kaladinlight, line 41: Sample defaults are meant to be minimal case for local kube dev. Is `5Gi` required by cluster or can we go with `1Gi`?

coinstacks/common/pulumi/src/rabbitmq.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export async function deployRabbit(
+  app: string,
+  asset: string,
+  provider: k8s.Provider,
+  namespace: string,
+  config: Pick<Config, 'rabbit'>
+  ): Promise<CustomResource | undefined> {
+
```
kaladinlight, line 22: There is typically a check `if (config.rabbit === undefined) return` to not deploy the rabbit cluster if not defined depending on config requirement.

coinstacks/common/pulumi/src/rabbitmq.ts
```diff
@@ -0,0 +1,106 @@
...
+  const clusterName = `${asset}-rabbitmq`
+
+  const resources: pulumi.Input<types.input.rabbitmq.v1beta1.RabbitmqClusterSpecResourcesArgs> = {
+    limits: {
+      cpu: config.rabbit?.cpuLimit as pulumi.Input<string>,
```
kaladinlight, line 27: Can strings not be fed into an input?

kaladinlight (approved):

coinstacks/common/pulumi/src/rabbitmq.ts
samortaylor: I think there might be a cleaner way to specify these values as `pulumi.Input`s, they cannot be strings.

PR #128, opened 2021-11-03
Transaction Parsing Module

### LINKED ISSUE #127: Parsing module
Discussion related to pulling out the parsing logic into it's own module to allow the parsing to happen client side. This will begin the work of modularizing the parsing logic in an effort to move towards client side parsing.

kaladinlight:
Start pulling parsing logic out into separate package and refine payload structure for easier client ingestion

```
      const expected: Tx = {
        txid: tx.txid,
        blockHeight: tx.blockHeight,
        blockTime: tx.blockTime,
        blockHash: tx.blockHash,
        address: address,
        caip2: 'ETH',
        value: tx.value,
        fee: {
          value: '78183644000000000',
          caip19: 'ETH',
        },
        transfers: [
          {
            type: TransferType.Send,
            from: '0xb8b19c048296E086DaF69F54d48dE2Da444dB047',
            to: '0x0d4a11d5EEaaC28EC3F61d100daF4d40471f1852',
            caip19: 'USDT',
            totalValue: '45000000000',
            components: [{ value: '45000000000' }],
            token: usdtToken,
          },
          {
            type: TransferType.Receive,
            from: '0xF82d8Ec196Fb0D56c6B82a8B1870F09502A49F88',
            to: '0xb8b19c048296E086DaF69F54d48dE2Da444dB047',
            caip19: 'KISHU',
            totalValue: '9248567698016204727450',
            components: [{ value: '9248567698016204727450' }],
            token: kishuToken,
          },
        ],
        trade: {
          dexName: Dex.Zrx,
          type: TradeType.Trade,
        },
      }
      ```

### CODE REVIEW

File: packages/parser/src/ethereum/thor.ts [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+
+  constructor(args: ParserArgs) {
+    this.abiInterface = new ethers.utils.Interface(ABI)
+    this.thorchain = new Thorchain({ midgardUrl: args.midgardUrl })
+
+    this.depositSigHash = this.abiInterface.getSighash('deposit')
+    this.transferOutSigHash = this.abiInterface.getSighash('transferOut')
+
+    // TODO: Router contract can change, use /inbound_addresses endpoint to determine current router contract
+    this.routerContract = {
```
0xdef1cafe, line 32: how is this going to work over time? won't we need to maintain a mapping of inbound_addresses over time?

kaladinlight: For this method, yes you are correct. And I am not sure if there is actually a way to get past router contracts if you weren't aware of it at the time.

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+  private zrx: zrx.Parser
+
+  constructor(args: TransactionParserArgs) {
+    const provider = new ethers.providers.JsonRpcProvider(args.rpcUrl)
+
+    this.network = args.network ?? Network.Mainnet
+
+    this.thor = new thor.Parser({ network: this.network, midgardUrl: args.midgardUrl })
+    this.uniV2 = new uniV2.Parser({ network: this.network, provider })
+    this.zrx = new zrx.Parser()
```
cjthompson, line 32: I think it would be better to have a `registerParser` mechanism to decouple the `ethereum` parser from these specific implementations. If we think of this package as independently useful, it would be great for developers to be able to write arbitrary parsers they can register rather than relying on the package to include support before they can use it.

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      case multiSig.SENDMULTISIG_SIG_HASH:
+        return multiSig.getInternalAddress(inputData)
+      default:
+        return
+    }
+  }
+
+  async parse(tx: Tx, address: string, internalTxs?: Array<InternalTx>): Promise<ParseTx> {
+    const sendAddress = tx.vin[0].addresses?.[0] ?? ''
+    const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
```
cjthompson, line 49: Do we need any level of validation to consider a transaction valid?

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      }
+      case this.thor.routerContract: {
+        result = await this.thor.parse(tx)
+        break
+      }
+      case uniV2.ROUTER_CONTRACT: {
+        result = await this.uniV2.parse(tx)
+        break
+      }
+    }
```
cjthompson, line 65: I would invert this logic and ask each parser if it wants to parse the tx, or you just call `parse` on all the available parsers (maybe stop at the first one that returns a result, or just accept all the results if we think that we want more than one parser to be able to process a transaction).

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      trade: result?.trade,
+      transfers: result?.transfers ?? [],
+      txid: tx.txid,
+      value: tx.value,
+    }
+
+    if (address === sendAddress) {
+      // send amount
+      const sendValue = new BigNumber(tx.value)
+      if (!sendValue.isNaN() && sendValue.gt(0)) {
```
cjthompson, line 82:
```suggestion
      if (sendValue.gt(0)) {
```
`sendValue.gt(0)` is false is `isNaN() === true`

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+          'ETH',
+          sendAddress,
+          receiveAddress,
+          sendValue.toString(10)
+        )
+      }
+
+      // network fee
+      const fees = new BigNumber(tx.fees ?? 0)
+      if (!fees.isNaN() && fees.gt(0)) {
```
cjthompson, line 95:
```suggestion
      const fees = new BigNumber(tx.fees)
      if (fees.gt(0)) {
```

The `?? 0` isn't accomplishing anything because you're checking if it's `> 0` so if it's `null | undefined`  it'll still be false.

```
> (new B(undefined)).gt(0)
false
> (new B(null)).gt(0)
false
> (new B(null || 0)).gt(0)
false
> (new B(undefined || 0)).gt(0)
false
```

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+
+    tx.tokenTransfers?.forEach((transfer) => {
+      // FTX Token (FTT) name and symbol was set backwards on the ERC20 contract
+      if (transfer.token == '0x50D1c9771902476076eCFc8B2A83Ad6b9355a4c9') {
+        transfer.name = transfer.symbol
+        transfer.symbol = transfer.name
+      }
+
```
cjthompson, line 121: This is mutating the `tx` object that's passed into the `parse` function. We should treat `tx` as immutable because the caller is probably not expecting their data to be mutated.

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    from: string,
+    to: string,
+    value: string,
+    token?: Token
+  ): Array<Transfer> {
+    const index = transfers?.findIndex((t) => t.type === type && t.caip19 === caip19 && t.from === from && t.to === to)
+    const transfer = transfers?.[index]
+
+    if (transfer) {
+      transfer.totalValue = new BigNumber(transfer.totalValue).plus(new BigNumber(value)).toString(10)
```
cjthompson, line 171: may result in `"NaN"`

packages/parser/src/ethereum/index.ts
```diff
@@ -0,0 +1,180 @@
...
+      transfer.components.push({ value: value })
+      transfers[index] = transfer
+    } else {
+      transfers = [...transfers, { type, caip19, from, to, totalValue: value, components: [{ value: value }], token }]
```
cjthompson, line 175:
```suggestion
      return [...transfers, { type, caip19, from, to, totalValue: value, components: [{ value: value }], token }]
```

it's preferable to not reassign function argument values. I would expect eslint/typescript to complain about that.

cjthompson:
This is a great start. I especially love the focus on tests to validate the complex logic.

I see a few areas where the architecture could be tweaked to make it more generic and allow other people to write additional parsers.

packages/parser/src/ethereum/index.ts
kaladinlight: Agreed that would be a nice improvement. I started this module as a pretty 1:1 swap out of the current ingester parser. We can start making issues to track enhancements as deemed necessary

packages/parser/src/ethereum/index.ts
kaladinlight: Can always log if there is a case where we don't have addresses to look into further. Blockbook is going to send back all transactions it finds in this format or return an error.

packages/parser/src/ethereum/index.ts
kaladinlight: I remember you pointing this out on a previous pr. Updating. I did not include 0 value as the parsed transaction just tracks information related to actual transfer of value. A 0 value send would imply a gas value for interacting with a contract or sending a token which we will catch.

packages/parser/src/ethereum/index.ts
kaladinlight: Removing isNaN check and I had added that because `undefined` is not of `type Value = string | number | Instance`.

packages/parser/src/ethereum/index.ts
kaladinlight: I would agree typically, but this is an actual bug in the erc20 contract itself and we have had to add it to both wt and unchained. The intention is to swap the symbol and name values to be assigned correctly for this known contract only.

packages/parser/src/ethereum/index.ts
kaladinlight: Good call

packages/parser/src/ethereum/index.ts
kaladinlight: No complaints with strict tsconfig rules and default eslint. For the existing transfer case (first if statement) we reassign the value of the transfer, do you suggest a different way for that as well?

kaladinlight:
> This is a great start. I especially love the focus on tests to validate the complex logic.
> 
> I see a few areas where the architecture could be tweaked to make it more generic and allow other people to write additional parsers.

Thanks and great feedback! I agree there is def room for improvement to make this more ergonomic and like the idea for the ability to inject custom parsers. Down to talk over those ideas more when we get the initial replacement swapped out for chain-adapters/web so we have caip and an easier payload to interact with!

packages/parser/src/ethereum/index.ts
kaladinlight: I like this concept. Another aspect for the continued improvement I will add as an issue to discuss and implement after initial cut.

PR #129, opened 2021-11-03
Update ethereum to return fee data for both legacy and eip-1559 support

### LINKED ISSUE #134: ETH: EIP-1559 support (fees)
Currently only `gasPrice` is being returned. Add `maxFeePerGas` and `maxPriorityFeePerGas` information for sending eip-1559 transactions as well.

kaladinlight:
Only `gasPrice` was being returned previously which is used for legacy transactions. Update to return fee data which includes `maxFeePerGas` and `maxPriorityFeePerGas` for EIP-1559 support.


cjthompson:
Can you please add a description

cjthompson:
This changes the response format of an API call (backwards incompatible). Can you make a corresponding PR to lib or web where this API is consumed?

kaladinlight:
> This changes the response format of an API call (backwards incompatible). Can you make a corresponding PR to lib or web where this API is consumed?

Yes I will take care of this, no package is being published or updated in lib yet so merging will not break.

cjthompson (approved):

PR #135, opened 2021-11-04
BTC: Websocket support

### LINKED ISSUE #133: BTC: websocket support

kaladinlight:


coinstacks/bitcoin/api/src/websocket.ts
```diff
@@ -0,0 +1,83 @@
+import WebSocket from 'isomorphic-ws'
+import { ErrorResponse, RequestPayload, Topics, TxsTopicData } from '@shapeshiftoss/common-api'
+import { SequencedBTCParseTx } from '@shapeshiftoss/bitcoin-ingester'
```
cjthompson, line 3: If you're only importing types and no code, you can use `import type`
```suggestion
import type { ErrorResponse, RequestPayload, Topics, TxsTopicData } from '@shapeshiftoss/common-api'
import type { SequencedBTCParseTx } from '@shapeshiftoss/bitcoin-ingester'
```

coinstacks/bitcoin/api/src/websocket.ts
```diff
@@ -0,0 +1,83 @@
...
+    ws.onclose = () => this.connections.txs?.pingTimeout && clearTimeout(this.connections.txs.pingTimeout)
+    ws.onmessage = (event) => {
+      if (event.data === 'ping') {
+        this.heartbeat('txs')
+        return
+      }
+
+      try {
+        const message = JSON.parse(event.data.toString()) as SequencedBTCParseTx | ErrorResponse
+        if ('type' in message) {
```
cjthompson, line 58: Can you add a comment explaining why you're checking for `type` here?

coinstacks/bitcoin/api/src/websocket.ts
```diff
@@ -0,0 +1,83 @@
...
+          onError && onError(message)
+          return
+        }
+        onMessage(message)
+      } catch (err) {
+        if (onError && err instanceof Error) onError({ type: 'error', message: err.message })
+      }
+    }
+
+    await new Promise((resolve) => (ws.onopen = () => this.onOpen('txs', resolve)))
```
cjthompson, line 68: I would suggest a `Promise.race` with a timeout to handle a case where a websocket cannot open, if such a scenario can happen.

coinstacks/bitcoin/api/src/websocket.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const payload: RequestPayload = { method: 'subscribe', data }
+    ws.send(JSON.stringify(payload))
+  }
+
+  unsubscribeTxs(): void {
+    this.connections.txs?.ws.send(
+      JSON.stringify({ method: 'unsubscribe', topic: 'txs', data: undefined } as RequestPayload)
```
cjthompson, line 76: casting is a code smell, meaning you're ignoring a problem the compiler is telling you. I don't even see why you need to cast in this case.

coinstacks/bitcoin/ingester/src/parseTx/index.ts
```diff
@@ -23,23 +23,23 @@ export const parseTx = async (tx: Tx, address: string): Promise<BTCParseTx> => {
...
+      }
 
-        // network fee (only for sends)
-        const fees = new BigNumber(tx.fees ?? 0)
-        if (!fees.isNaN() && fees.gt(0)) {
-          pTx.fee = { symbol: 'BTC', value: fees.toString(10) }
-        }
+      // network fee
+      const fees = new BigNumber(tx.fees ?? 0)
+      if (!fees.isNaN() && fees.gt(0)) {
```
cjthompson, line 36: `fees.gt(0)` will return `false` if `fees.isNaN()` is true, so you don't need to do both checks.

in web we use this helper:
```ts
export const bnOrZero = (n: BigNumber.Value | null | undefined): BigNumber => {
  const value = new BigNumber(n || 0)
  return value.isNaN() ? bn(0) : value
}
```

cjthompson (approved):
Approved with comments

coinstacks/bitcoin/api/src/websocket.ts
kaladinlight: sure!

coinstacks/bitcoin/api/src/websocket.ts
kaladinlight: ws is all json, I just wanted to add the cast to ensure if types change it would catch that the desired json payload may no longer match.

coinstacks/bitcoin/api/src/websocket.ts
kaladinlight: Will make a separate issue for this

PR #146, opened 2021-11-15
RabbitMQ Memory

samortaylor:
We have been experiencing issues with ingesters communicating with RabbitMQ, memory alerts have been found.  

-Double RabbitMQ memory
-Increase logging verbosity on coinstacks


kaladinlight (approved):
:fire:

PR #160, opened 2021-11-30
Add caip standard

kaladinlight:


yarn.lock
```diff
@@ -2182,6 +2189,16 @@
...
+"@shapeshiftoss/types@^1.11.4":
+  version "1.13.1"
+  resolved "https://registry.yarnpkg.com/@shapeshiftoss/types/-/types-1.13.1.tgz#260a17f6a9555ab78fd637049ef10b3e6c2d0fd0"
+  integrity sha512-krddcUTaJSVCI2EIXzLS61siTj5RDJUvKqu6/LYXNrDjML4uoHp5PQ4NUFrTT+p0K6tJ2oMLxV4FwSK7maZkEA==
+
+"@shapeshiftoss/types@^1.14.0":
+  version "1.14.0"
+  resolved "https://registry.yarnpkg.com/@shapeshiftoss/types/-/types-1.14.0.tgz#d138f68105b403555f2a58141a050bfe2f03eb23"
+  integrity sha512-mL9ynAEzovGZ1sTrvDr19nuIe0GOabilU0ofoSjGMstyHL130mjmyQYsx5/MdP8ldJj9irAuWHkwuGDFaRHXQw==
+
```
cjthompson, line 2201: Why are there two of these?

jest.config.js
```diff
@@ -5,7 +5,9 @@ module.exports = {
   testEnvironment: 'node',
   testPathIgnorePatterns: ['.d.ts', '.js', '__mocks__'],
   moduleNameMapper: {
+    '^@shapeshiftoss\\/caip': '@shapeshiftoss/caip',
     '^@shapeshiftoss\\/logger': '@shapeshiftoss/logger',
+    '^@shapeshiftoss\\/types': '@shapeshiftoss/types',
     '^@shapeshiftoss\\/([^/]+)': '@shapeshiftoss/$1/src',
```
cjthompson, line 11: My recommendation would be to use a different prefix for internal packages

cjthompson (approved):

yarn.lock
kaladinlight: Two different versions being referenced in lib. Are you saying it is now going to be a requirement for any project import lib packages to run `yarn-minify` because of versioning issues in that repo?

jest.config.js
kaladinlight: Everything is technically being published at this point, so the `shapeshiftoss` scope name is required. The module mapping was done specifically so the `__mocks__` could be referenced correctly, so maybe there is a more correct way to do that piece to clean this up.

kaladinlight: Found a better way to handle both cases: `'^@shapeshiftoss\\/([^/]+)': ['@shapeshiftoss/$1/src', '@shapeshiftoss/$1'],`

PR #210, opened 2022-01-11
Add go-swagger annotations to /info

kaladinlight:
- Adds go-swagger annotations for generating swagger spec
- Includes make target for generation using docker
- Defines same error responses as ts api
- Update base interface to return interfaces from functions to allow extending while still implementing the BaseAPI interface
- Add interface compile check to ensure coinstack properly implements BaseAPI interface


asamere (approved):

alexanderramos89 (commented):
go/coinstacks/cosmos/Makefile

PR #220, opened 2022-01-24
fix up osmosis and thorchain handlers

asamere:


go/coinstacks/osmosis/api/handlers.go
```diff
@@ -12,7 +11,10 @@ type Handler struct {
 }
 
 func (h *Handler) GetInfo() (api.Info, error) {
-	return api.Info{Network: "mainnet"}, nil
+	info := api.BaseInfo{
```
kaladinlight, line 14: Use osmosis.Info type which has BaseInfo nested.

kaladinlight (approved):
One comment, rest looks good

.gitignore
```diff
@@ -0,0 +1 @@
+go/.vscode/launch.json
```
kaladinlight, line 1: Also, we have `.gitignore` in `unchained/go` that we can add this

PR #233, opened 2022-02-01
unchained client - generate cosmos 

pastaghost:


go/Makefile
```diff
@@ -1,5 +1,5 @@
 COINSTACKS := $(wildcard coinstacks/*)
-MAKEFLAGS += -j$(words $(COINSTACKS)) --no-print-directory
```
kaladinlight, line 2: Can we not keep these jobs in parallel by increasing the resources for docker desktop?

kaladinlight (approved):
One comment, but the rest looks good.

go/Makefile
pastaghost: We can, I just figured that it might be better to run the jobs sequentially since external contributors may run into the same problem that I did and it isn't obvious what the source of the error is. It's your call on this one, but I was thinking that since the time difference between the sequential and parallel builds isn't substantial, it might be better to play it safe.

kaladinlight: Yeah fair enough. The `$(words $COINSTACKS))` can be removed along with the flag.

kaladinlight:
It also looks like we need to update circle to generate the swagger in the go coinstacks. Let me take a look at that.

Issue #236, opened 2022-02-05
Move custom transaction parsing logic from web to unchained


0xApotheosis:
## Description
Move the "messy" logic in `web`'s `useTxDetails.ts` custom hook upstream to the parsing logic in `unchained`.

## Work to be done
Support a standard transaction with the new backend logic that meet feature parity with the current front-end logic.

PR #238, opened 2022-02-07
refactor: transaction parsing logic

0xApotheosis:
## Description

Refactor the transaction parsing logic. Each parser now handles its own logic to determine if it supports a transaction.

All parsers are checked for a response, and the first not to return `undefined` is used to get a parsed response.

Intended to support https://github.com/shapeshift/unchained/issues/236.

### DIFF
--- a/node/packages/parser/src/bitcoin/index.ts
+++ b/node/packages/parser/src/bitcoin/index.ts
@@ -8,7 +8,7 @@ import { toNetworkType } from './utils'
 
 export * from './types'
 
-export interface TransactionParserArgs {
+export interface BitcoinTransactionParserArgs {
   network?: Network
   rpcUrl: string
 }
@@ -16,7 +16,7 @@ export interface TransactionParserArgs {
 export class TransactionParser {
   network: Network
 
-  constructor(args: TransactionParserArgs) {
+  constructor(args: BitcoinTransactionParserArgs) {
     this.network = args.network ?? 'mainnet'
   }
 
--- a/node/packages/parser/src/bitcoin/index.ts
+++ b/node/packages/parser/src/bitcoin/index.ts
@@ -8,7 +8,7 @@ import { toNetworkType } from './utils'
 
 export * from './types'
 
-export interface BitcoinTransactionParserArgs {
+export interface TransactionParserArgs {
   network?: Network
   rpcUrl: string
 }
@@ -16,7 +16,7 @@ export interface BitcoinTransactionParserArgs {
 export class TransactionParser {
   network: Network
 
-  constructor(args: BitcoinTransactionParserArgs) {
+  constructor(args: TransactionParserArgs) {
     this.network = args.network ?? 'mainnet'
   }
 
--- a/node/packages/parser/src/ethereum/helpers.ts
+++ b/node/packages/parser/src/ethereum/helpers.ts
@@ -4,15 +4,3 @@ export const txInteractsWithContract = (tx: Tx, contract: string) => {
   const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
   return receiveAddress === contract
 }
-
-export async function findAsyncSequential<T>(
-  array: T[],
-  predicate: (element: T) => Promise<boolean>
-): Promise<T | undefined> {
-  for (const element of array) {
-    if (await predicate(element)) {
-      return element
-    }
-  }
-  return undefined
-}
--- a/node/packages/parser/src/ethereum/helpers.ts
+++ b/node/packages/parser/src/ethereum/helpers.ts
@@ -1,0 +1,10 @@
+import { EthereumSpecific, Tx } from '@shapeshiftoss/blockbook'
+
+export const txMatchesContract = (tx: Tx, contract: string) => {
+  const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
+  return receiveAddress === contract
+}
+
+export const isEthereumSpecific = (
+  maybeEthereumSpecific: EthereumSpecific | undefined
+): maybeEthereumSpecific is EthereumSpecific => !!maybeEthereumSpecific
--- a/node/packages/parser/src/ethereum/helpers.ts
+++ b/node/packages/parser/src/ethereum/helpers.ts
@@ -1,10 +1,18 @@
-import { EthereumSpecific, Tx } from '@shapeshiftoss/blockbook'
+import { Tx } from '@shapeshiftoss/blockbook'
 
-export const txMatchesContract = (tx: Tx, contract: string) => {
+export const txInteractsWithContract = (tx: Tx, contract: string) => {
   const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
   return receiveAddress === contract
 }
 
-export const isEthereumSpecific = (
-  maybeEthereumSpecific: EthereumSpecific | undefined
-): maybeEthereumSpecific is EthereumSpecific => !!maybeEthereumSpecific
+export async function findAsyncSequential<T>(
+  array: T[],
+  predicate: (element: T) => Promise<boolean>
+): Promise<T | undefined> {
+  for (const element of array) {
+    if (await predicate(element)) {
+      return element
+    }
+  }
+  return undefined
+}
--- a/node/packages/parser/src/ethereum/index.ts
+++ b/node/packages/parser/src/ethereum/index.ts
@@ -10,7 +10,7 @@ import * as multiSig from './multiSig'
 import * as thor from './thor'
 import * as uniV2 from './uniV2'
 import * as zrx from './zrx'
-import { findAsyncSequential } from './helpers'
+import { findAsyncSequential } from '../helpers'
 
 export * from './types'
 
@@ -59,7 +59,7 @@ export class TransactionParser {
     const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
 
     // We expect only one Parser to return a result. If multiple do, we take the first and early exit.
-    const parser = await findAsyncSequential(this.parsers, async (parser) => !!(await parser.parse(tx)))
+    const parser = await findAsyncSequential(this.parsers, async (parser) => await parser.parse(tx))
     const result = await parser?.parse(tx)
 
     const pTx: ParseTx = {
--- a/node/packages/parser/src/ethereum/index.ts
+++ b/node/packages/parser/src/ethereum/index.ts
@@ -13,7 +13,7 @@ import * as zrx from './zrx'
 
 export * from './types'
 
-export interface EthereumTransactionParserArgs {
+export interface TransactionParserArgs {
   network?: Network
   midgardUrl: string
   rpcUrl: string
@@ -26,7 +26,7 @@ export class TransactionParser {
   private uniV2: uniV2.Parser
   private zrx: zrx.Parser
 
-  constructor(args: EthereumTransactionParserArgs) {
+  constructor(args: TransactionParserArgs) {
     const provider = new ethers.providers.JsonRpcProvider(args.rpcUrl)
 
     this.network = args.network ?? 'mainnet'
--- a/node/packages/parser/src/ethereum/index.ts
+++ b/node/packages/parser/src/ethereum/index.ts
@@ -59,8 +59,7 @@ export class TransactionParser {
     const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
 
     // We expect only one Parser to return a result. If multiple do, we take the first and early exit.
-    const parser = await findAsyncSequential(this.parsers, async (parser) => await parser.parse(tx))
-    const result = await parser?.parse(tx)
+    const result = await findAsyncSequential(this.parsers, async (parser) => await parser.parse(tx))
 
     const pTx: ParseTx = {
       address,
--- a/node/packages/parser/src/ethereum/index.ts
+++ b/node/packages/parser/src/ethereum/index.ts
@@ -10,6 +10,7 @@ import * as multiSig from './multiSig'
 import * as thor from './thor'
 import * as uniV2 from './uniV2'
 import * as zrx from './zrx'
+import { findAsyncSequential } from './helpers'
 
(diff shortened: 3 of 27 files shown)


node/packages/parser/src/bitcoin/index.ts
```diff
@@ -8,15 +8,15 @@ import { toNetworkType } from './utils'
 
 export * from './types'
 
-export interface TransactionParserArgs {
```
kaladinlight, line 11: As these classes are all scoped within coin specific packages and we export them as named vars (ie. `export * as ethereum`), I intentionally did not prefix with the coin name again. I have always felt like this is redundant naming that we have through a lot of our stack (even though the majority of packages are separated by coin), but maybe that is not a shared opinion. Just wanted to provide a bit of context. Let me know your thoughts!

0xApotheosis: That's useful context, thanks!

I don't think there is a right answer for this, though my thought process here is the change is a slight net improvement, even considering the redundancy/more verbose naming issues, which I agree with.

What I observe - although the Typescript compiler has no issues, IDEs can struggle in instances like this. As an example, when I use WebStorm to check the type of `TransactionParser` in https://github.com/shapeshift/unchained/blob/d6909f769cc943300ffeb7370e9f4a2b546a0da8/node/coinstacks/ethereum/ingester/src/workers/address.ts#L20, it actually resolves to `bitcoin`'s `TransactionParser`. 

I'm curious to know if you have the same issue in your IDE? If not, it could just be poor IDE configuration on my behalf, in which case we are muddying the codebase for no reason, and we should absolutely revert his change.

In summary, I agree - they are isolated modules that have no knowledge of each other, and _shouldn't_ need a domain prefix.

0xApotheosis: As far as I can tell it's a WebStorm / JetBrain issue, where it'll resolve all types that match the name and return the first one.

In that case, I'll revert the changes in this commit, as we shouldn't compromise our code standards to work around a JetBrains bug.

node/packages/parser/src/ethereum/index.ts
```diff
@@ -52,21 +57,9 @@ export class TransactionParser {
...
-      case this.thor.routerContract: {
-        result = await this.thor.parse(tx)
-        break
-      }
-      case uniV2.ROUTER_CONTRACT: {
-        result = await this.uniV2.parse(tx)
-        break
-      }
-    }
+    const parserResults = await Promise.all(this.parsers.map(async (parser) => await parser.parse(tx)))
```
kaladinlight, line 60: Might be worth using a `for await of` loop so we can just break out of the loop on first successful parse

node/packages/parser/src/ethereum/thor.ts
```diff
@@ -48,21 +49,28 @@ export class Parser {
...
-      case this.depositSigHash:
-        result = this.abiInterface.decodeFunctionData(this.depositSigHash, tx.ethereumSpecific.data)
-        break
-      case this.transferOutSigHash: {
-        result = this.abiInterface.decodeFunctionData(this.transferOutSigHash, tx.ethereumSpecific.data)
-        break
+    if (!this.isSupportedTransaction(tx)) return
+    if (!(isEthereumSpecific(tx.ethereumSpecific) && tx.ethereumSpecific.data)) return
+
+    const result: ethers.utils.Result | undefined = (() => {
```
kaladinlight, line 60: Down to make use of anon inline functions to avoid `let`. Is the type definition required, or is the implicit type assigned by default?

node/packages/parser/src/ethereum/index.ts
```diff
@@ -19,12 +19,15 @@ export interface TransactionParserArgs {
...
 export class TransactionParser {
   network: Network
 
-  private thor: thor.Parser
-  private uniV2: uniV2.Parser
-  private zrx: zrx.Parser
+  private readonly thor: thor.Parser
+  private readonly uniV2: uniV2.Parser
+  private readonly zrx: zrx.Parser
+  private readonly parsers: Array<Parser>
```
kaladinlight, line 30: We probably don't need to keep track of these individually anymore.

node/packages/parser/src/ethereum/index.ts
```diff
@@ -19,12 +19,15 @@ export interface TransactionParserArgs {
...
+
 export class TransactionParser {
   network: Network
 
-  private thor: thor.Parser
-  private uniV2: uniV2.Parser
-  private zrx: zrx.Parser
+  private readonly thor: thor.Parser
+  private readonly uniV2: uniV2.Parser
+  private readonly zrx: zrx.Parser
```
kaladinlight, line 29: We probably don't need to keep track of these individually anymore.

node/packages/parser/src/ethereum/index.ts
```diff
@@ -19,12 +19,15 @@ export interface TransactionParserArgs {
   rpcUrl: string
 }
 
+export type Parser = thor.Parser | uniV2.Parser | zrx.Parser
```
kaladinlight, line 22: We should think about making an interface that we can use and conform to for these parsers, especially if we are going to support custom parser injection. Doesn't need to be included in this PR.

node/packages/parser/src/ethereum/zrx.ts
```diff
@@ -1,11 +1,18 @@
 import { Tx } from '@shapeshiftoss/blockbook'
-import { Dex, Tx as ParseTx, TradeType } from '../types'
+import { Dex, TradeType, Tx as ParseTx } from '../types'
+import { txMatchesContract } from './helpers'
 
 export const PROXY_CONTRACT = '0xDef1C0ded9bec7F1a1670819833240f027b25EfF'
 
 export class Parser {
+  isSupportedTransaction(tx: Tx) {
```
kaladinlight, line 8: You like the sounds of `isParsable` better?

kaladinlight: Can see this being a required part of the interface

node/packages/parser/src/ethereum/helpers.ts
```diff
@@ -0,0 +1,10 @@
+import { EthereumSpecific, Tx } from '@shapeshiftoss/blockbook'
+
+export const txMatchesContract = (tx: Tx, contract: string) => {
+  const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
+  return receiveAddress === contract
+}
+
+export const isEthereumSpecific = (
```
kaladinlight, line 8: I can get behind the `txMatchesContract` though maybe a name like `txInteractsWithContract` or something like that would be a little better. I don't quite understand the `isEthereumSpecific` helper and feels like it complicates things instead of just having a optional check inline `!tx.ethereumSpecific?.data`

node/packages/parser/src/ethereum/uniV2.ts
```diff
@@ -37,9 +38,15 @@ export class Parser {
...
+    const isUniV2Contract = txMatchesContract(tx, ROUTER_CONTRACT)
+    const hasNoConfirmations = tx.confirmations === 0
+    return hasNoConfirmations && isUniV2Contract
+  }
+
   async parse(tx: Tx): Promise<ParseTxSpecific | undefined> {
-    if (!tx.ethereumSpecific?.data) return
-    if (tx.confirmations !== 0) return
+    if (!this.isSupportedTransaction(tx)) return
+    if (!(isEthereumSpecific(tx.ethereumSpecific) && tx.ethereumSpecific.data)) return
```
kaladinlight, line 49: can just move this into the parsable/supported check to keep all of that logic together

node/packages/parser/src/ethereum/index.ts
0xApotheosis: I like where your head is at. I wrote a little find helper that lets us check the result of asynchronous calls, `findAsyncSequential`, that achieves this in a (subjectively) slightly cleaner way - it'll early exit with the first parser to return a result that is not `undefined`.

Commit https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599.

node/packages/parser/src/ethereum/thor.ts
0xApotheosis: Implicitly assigned - removed in https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599.

node/packages/parser/src/ethereum/index.ts
0xApotheosis: We do still use `this.thor` in `getInternalAddress`, and it felt a little weird keeping that property on the `TransactionParser` class, but removing the rest. Thoughts?

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Agree. I'll do another PR on this 👌

node/packages/parser/src/ethereum/zrx.ts
0xApotheosis: Definitely, `isParsable` better conveys its relationship with `parse()`.

Renamed in https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599

node/packages/parser/src/ethereum/zrx.ts
```diff
@@ -1,11 +1,18 @@
...
+    const isZrxContract = txMatchesContract(tx, PROXY_CONTRACT)
```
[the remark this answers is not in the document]
0xApotheosis: Good observation, I like the early return on each case - chaining `&&`s can get ugly fast. 

I'm hoping to find the right balance between concision and clarity for outside devs to contribute - though intermediary values are probably unnecessarily for the complexity of these. 

Hopefully I've found that balance in https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599.

node/packages/parser/src/ethereum/thor.ts
```diff
@@ -48,21 +49,28 @@ export class Parser {
     return result.to
   }
 
+  isSupportedTransaction(tx: Tx) {
```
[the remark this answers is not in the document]
0xApotheosis: As discussed on Discord, I added explicit return types in https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599 - I won't spend time looking at the lint issue for now.

node/packages/parser/src/ethereum/helpers.ts
0xApotheosis: Updated function name to `txInteractsWithContract` in https://github.com/shapeshift/unchained/pull/238/commits/143f663f3aa13c278ce57bef914db30cc1be0599.

You're right about the Type Guard, that was unnecessarily complicated in this case.

node/packages/parser/src/ethereum/uniV2.ts
0xApotheosis: Unfortunately, the compiler isn't smart enough to realise that `tx.ethereumSpecific.data` is defined if we do the check inside `isParsable`, so we then need to do a bunch of checks in `parse()`.

I'm not aware of a way around this - am I missing something?

node/packages/parser/src/ethereum/index.ts
kaladinlight: Oh yeah, you are right. This is currently being used in the ingester which will be ripped out at some point, but until then let's just leave it as is.

node/packages/parser/src/ethereum/uniV2.ts
kaladinlight: No you are correct, we do lose the type narrowing scope by moving that into `isParsable`. Avoiding additional undefined checks is the right call.

kaladinlight: The only other option would be just forgoing the `isParsable` function and just having the few checks at the start of the `parse` function.

node/packages/parser/src/ethereum/index.ts
```diff
@@ -52,21 +58,9 @@ export class TransactionParser {
...
-        result = await this.thor.parse(tx)
-        break
-      }
-      case uniV2.ROUTER_CONTRACT: {
-        result = await this.uniV2.parse(tx)
-        break
-      }
-    }
+    // We expect only one Parser to return a result. If multiple do, we take the first and early exit.
+    const parser = await findAsyncSequential(this.parsers, async (parser) => !!(await parser.parse(tx)))
```
kaladinlight, line 62: It looks like this will run through all parsers, return on the first one that parsed the transaction, and then run `await parser?.parse(tx)` again. Can we just have the for of loop return the first result we get instead?

That function also seems general enough to have it as a top level parser helper instead of an ethereum helper specifically if we want to keep it as a helper function.

node/packages/parser/src/ethereum/index.ts
kaladinlight: It looks like this will run through all parsers, return on the first one that parsed the transaction, and then run await parser?.parse(tx) again. Can we just have the for of loop return the first result we get instead?

That function also seems general enough to have it as a top level parser helper instead of an ethereum helper specifically if we want to keep it as a helper function.

0xApotheosis: Indeed, it was running the `parse()` twice. Luckily idempotent, but a waste of compute.

Fixed in https://github.com/shapeshift/unchained/pull/238/commits/78748721415c9c0242fcb0842bee3b6b7a10b1a7.

node/packages/parser/src/ethereum/uniV2.ts
0xApotheosis: Yeah, the checks probably aren't even worth splitting into `isParsable()` - I've removed all instances of it for now in https://github.com/shapeshift/unchained/pull/238/commits/68fc4e8bdfc87f4198611e671ec694da193f39d4.

kaladinlight (approved):

PR #247, opened 2022-02-10
feat: add buyTx and sellTx logic to unchained

0xApotheosis:
## Description

Move some logic (`buyTx` & `sellTx`) in `web`'s `useTxDetails.ts` custom hook upstream to the parsing logic in `unchained`.

## Discussion

I'm wary of the temporal coupling we've got going on in the `parser()` in `ethereum/parser` because of the mutations to `pTx`.

E.g. attempting to use `pTx` to calculate the `metaData` too early will return incorrect data, as transactions etc might not yet be added.

@kaladinlight if you're open to it I'd be keen to do a follow up PR to address this.


node/packages/parser/src/ethereum/index.ts
```diff
@@ -59,7 +60,10 @@ export class TransactionParser {
     const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
 
     // We expect only one Parser to return a result. If multiple do, we take the first and early exit.
-    const result = await findAsyncSequential(this.parsers, async (parser) => await parser.parse(tx))
+    const contractParserResult: TxSpecific<ParseTx> | undefined = await findAsyncSequential(
```
kaladinlight, line 63: If we update the return type of the predicate parameter in `findAsyncSequential` to `Promise<U | undefined>` to match the signature of contract `parse()`, we can then utilize the generic types and get the correct return type without needing to cast `findAsyncSequential<Parser, TxSpecific<ParseTx>>()`

node/packages/parser/src/types.ts
```diff
@@ -8,7 +8,7 @@ export interface Fee {
   value: string
 }
 
-export interface SequencedTx extends Tx {
+export interface SequencedTx extends TxBase {
```
kaladinlight, line 11: Shouldn't this still extend <`TxBase` | `Tx`> as the sequenced payload we are going to be ingesting client side is possibly a `UniV2Tx` or `ZrxTx` or `ThorTx` as well? Unless we can just add `TxBase` to the `Tx` union

```suggestion
export interface SequencedTx extends <TxBase | Tx> {
```

node/packages/parser/src/ethereum/index.ts
```diff
@@ -171,6 +175,19 @@ export class TransactionParser {
       }
     })
 
+    const metaData: TxMetadata = (() => {
```
kaladinlight, line 178: I was initially thinking the contract parsers would return possible tx metadata depending on what it saw via contract interaction. If we went this route and made use of the input data to determine interaction, we would be able to uncouple the metadata from the parsed transfers. Not only that, but it seems like we would need the flexibility to do more custom metadata parsing that would have to belong in the contract parsers regardless (though maybe this is what you are implying with the follow up PR).

kaladinlight: I am also not a fan of this anon function. It is just a more verbose creation of an object versus:
```
const metaData: TxMetaData = {
  buyTx: getBuyTx(pTx),
  sellTx: getSellTx(pTx)
}
```

node/packages/parser/src/ethereum/index.ts
```diff
@@ -171,6 +175,19 @@ export class TransactionParser {
...
+      const buyTx = getBuyTx(pTx)
+      const sellTx = getSellTx(pTx)
+      return {
+        buyTx,
+        sellTx,
+      }
+    })()
+
+    if (metaData) {
```
kaladinlight, line 187: This will always evaluate to true as it stands

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Very clean. Fixed in https://github.com/shapeshift/unchained/pull/247/commits/0b9cbee47744a18398c252dc10bcda1a33989f92.

node/packages/parser/src/types.ts
0xApotheosis: You're right, though as we can't extend a union type using an interface, I think the best way to handle it is with a type alias, which I've committed in https://github.com/shapeshift/unchained/pull/247/commits/0b9cbee47744a18398c252dc10bcda1a33989f92:

```ts
export type SequencedTx = Tx & {
  sequence: number
  total: number
}
```

kaladinlight: Good point. I typically prefer interface (guess it would have been `extends TxBase, ZrxTx, ...`), but for sake of not needing to update two locations with the set of types, that works perfect. Now I think we still want to add `BaseTx` to the `Tx` union. If no contract parsers run, it would just be a `BaseTx` or standard tx returned

kaladinlight: Just noticed naming as well, for my ocd, can we change `TxBase` -> `BaseTx` (or what do you think of `StandardTx`?) so it matches `ZrxTx`, `ThorTx`, etc format.

0xApotheosis: Yep, `TxBase` bugged me too, the order of specificity was inverted. `StandardTx` 👌 https://github.com/shapeshift/unchained/pull/247/commits/d61821aabe4f0fc14c11e6c0671c0918625f359b

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Indeed, I got ahead of myself: https://github.com/shapeshift/unchained/pull/247/commits/f033e0557e89e877934340067a572504c431f00f

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Simplified in https://github.com/shapeshift/unchained/pull/247/commits/f033e0557e89e877934340067a572504c431f00f

I believe we are aligned re: the contract interaction determining the metadata. What I wanted to do first was just copy all the logic across from `web` in a naive way, and then get `web` using the transaction `data` it received from `unchained`.

Mitigates risk, and lets us firm up tests against existing logic (albeit in a new repo) before making bigger refactors.

kaladinlight (approved):

PR #255, opened 2022-02-14
tx history

asamere:
This was a tricky endeavor as cosmos nodes don't provide a means to query all txs for a given address. Instead you have to make 2 requests, one where the address created the tx (sender) and one where the address was affected by a tx created by another account (recipient). Sometimes an account is both sender and recipient so de-duplication is also required.

The tendermint RPC tx_search supports a block height query criteria, however performance was degraded by a good amount when using the block height in the tendermint query. In order to keep tx history performant, the last read page number, last tx hash, and total number of results is stored in an opaque (base64) cursor for sender and recipient txs.


go/cmd/cosmos/main.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	GRPCURL string `mapstructure:"grpcUrl"`
 	LCDURL  string `mapstructure:"lcdUrl"`
 	RPCURL  string `mapstructure:"rpcUrl"`
+	Scheme  string `mapstructure:"scheme"`
```
asamere, line 26: added Scheme as cosmos nodes are http by default, datahub is https

go/coinstacks/cosmos/api/handlers.go
```diff
@@ -69,9 +69,10 @@ func (h *Handler) GetAccount(pubkey string) (api.Account, error) {
 	return account, nil
 }
 
-func (h *Handler) GetTxHistory(pubkey string, page int, pageSize int) (api.TxHistory, error) {
-	res, err := h.httpClient.GetTxHistory(pubkey, page, pageSize)
+func (h *Handler) GetTxHistory(pubkey string, cursor string, pageSize uint) (api.TxHistory, error) {
+	res, err := h.httpClient.GetTxHistory(pubkey, cursor, pageSize)
 	if err != nil {
+		logger.Errorf("error getting tx history for %s (%s): %s", pubkey, cursor, err)
```
kaladinlight, line 75: Prefer wrapping errors with annotations on only logging at the top level caller

go/pkg/api/api.go
```diff
@@ -77,12 +77,9 @@ func (b BaseAccount) pubkey() string {
 // Contains info about pagination for large sets of data
 // swagger:model Pagination
 type Pagination struct {
-	// required: true
-	// example: 1
-	Page int `json:"page"`
-	// required: true
-	// example: 10
-	TotalPages int `json:"totalPages"`
+	// required: false
```
kaladinlight, line 80: Wouldn't the cursor be required?

go/pkg/cosmos/types.go
```diff
@@ -96,7 +96,8 @@ type Tx struct {
 }
 
 type TxHistory struct {
-	TotalPages int
+	TotalPages int // Deprecated
```
kaladinlight, line 99: We can just remove since tx history endpoint has never been used/fully completed

go/pkg/api/api.go
cjthompson: > Wouldn't the cursor be required?

No. If `cursor` is not set, then the caller is asking for the first page.

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+	return fmt.Sprintf(queryTemplate, address)
+}
+
+func buildRecvdTxsQuery(address string) string {
+	queryTemplate := `"transfer.recipient='%s'"`
+	return fmt.Sprintf(queryTemplate, address)
+}
+
+// given 2 sets of tx history, remove duplicates from the second set (recipient txs)
+func uniqueify(sender []client.TxSearchResponseResultTxs, recip []client.TxSearchResponseResultTxs) ([]client.TxSearchResponseResultTxs, []client.TxSearchResponseResultTxs) {
```
cjthompson, line 301: According to Google, `uniquify` is the accepted spelling

go/pkg/api/api.go
kaladinlight: This struct is for the return response, but I guess we might be using empty cursor as a way to say no more data instead of a `hasMore` key or something?

kaladinlight: Not sure what the best practice is for cursor based.

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+
+	// iterate until we get some results after uniquifying or run out of txs from tendermint
+	for {
+		start := time.Now().UnixMilli()
+		res, httpRes, err = s.tendermintClient.InfoApi.TxSearch(s.ctx).
+			Query(query).
+			Page(int32(page)).
+			PerPage(int32(tendermintPageSize)).
+			OrderBy("\"desc\"").Execute()
+		logger.Infof("type %d page %d query resolved in %dms: %s", txType, page, time.Now().UnixMilli()-start, query)
```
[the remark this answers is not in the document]
kaladinlight: Yes, all standard levels supported

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
-		if _, ok := seen[*t.Hash]; !ok {
-			uniqueTxs = append(uniqueTxs, t)
-			seen[*t.Hash] = true
+// read tendermint tx history
+func (s *txHistoryState) readTxHistory() (*TxHistory, error) {
+	// if we have cursor, restore state
+	if s.cursorSerialized != "" {
+		if err := s.cursor.decode(s.cursorSerialized); err != nil {
+			return nil, errors.Wrapf(err, "error decoding cursor %s", s.cursorSerialized)
+		} else {
```
kaladinlight, line 346: No need for else

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+			}
 		}
 	}
 
-	// sort descending by block height
-	sort.Slice(uniqueTxs, func(i, j int) bool {
-		if *uniqueTxs[i].Height == *uniqueTxs[j].Height {
-			return *uniqueTxs[i].Index > *uniqueTxs[j].Index
+	// default pageSize to 10
+	if s.pageSize == 0 {
```
kaladinlight, line 358: This check is already being handled at the top level handle func

go/pkg/cosmos/txhistory.go
```diff
@@ -19,6 +22,39 @@ import (
...
+	encodingConfig   *params.EncodingConfig
+	tendermintClient *client.APIClient
+}
+
+type HistTxType int
+
+const (
+	histTxTypeSender = iota
+	histTxTypeRecipient
+)
```
kaladinlight, line 54: Additional changes needed, but this would be a more idomatic representation of enum (typed iota, enum naming)
```suggestion
type TxType int

const (
	Sender TxType = iota
	Recipient
)
```

go/pkg/cosmos/txhistory.go
```diff
@@ -19,6 +22,39 @@ import (
 	"github.com/shapeshift/go-unchained/pkg/tendermint/client"
 )
 
+// cursor stores state client side between paginated requests
+type Cursor struct {
+	LastBlockHeight     uint64 `json:"blockHeight"`
```
kaladinlight, line 27: Why not just use `int` for number representations generally speaking?

go/cmd/cosmos/main.go
kaladinlight: thinking it might be best to update our config to include fully qualified uri string and we can `url.Parse()` where needed to pluck out scheme/port/etc

go/pkg/cosmos/txhistory.go
```diff
@@ -19,6 +22,39 @@ import (
...
+
+// primary history state struct
+type txHistoryState struct {
+	ctx              context.Context
+	address          string
+	pageSize         uint
+	cursor           Cursor
+	cursorSerialized string
+	sends            []client.TxSearchResponseResultTxs
+	recips           []client.TxSearchResponseResultTxs
```
kaladinlight, line 44: Would like to see same verbiage throughout. `sendTxs`, `receiveTxs`, `SendPage`, `ReceivePage`, `TotalReceiveTxs`, `TotalSendTxs`, etc

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
-	seen := make(map[string]bool)
-	uniqueTxs := []client.TxSearchResponseResultTxs{}
-	for _, t := range resTxs {
-		if _, ok := seen[*t.Hash]; !ok {
-			uniqueTxs = append(uniqueTxs, t)
-			seen[*t.Hash] = true
+// read tendermint tx history
+func (s *txHistoryState) readTxHistory() (*TxHistory, error) {
+	// if we have cursor, restore state
+	if s.cursorSerialized != "" {
```
kaladinlight, line 343: I would pull this logic out into `GetTxHistory`, we have the serialized cursor at that point that we can either decode or instantiate a new one to interact with throughout the rest of the logic. Then we don't need to have `cursorSerialized` in the struct either since we never use it for anything afterwards

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+		if err := s.cursor.decode(s.cursorSerialized); err != nil {
+			return nil, errors.Wrapf(err, "error decoding cursor %s", s.cursorSerialized)
+		} else {
+			// decrement by 1 so we start on last page we requested from tendermint
+			if s.cursor.TendermintSentPage > 0 {
+				s.cursor.TendermintSentPage--
```
kaladinlight, line 349: This seems weird to me and I would like to see whoever is calling or iterating the pages manage changing the page num. This is already a very complicated mutation state situation so making it as clear as possible who is changing the state and for what reason is going to go a long way to make things clear.

This does bring up the topic of even relying on such a heavily mutating state struct or if there would be a more functional way to accomplish this logic.

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+		tx := recip[i]
+		if _, ok := m[*tx.Hash]; !ok {
+			m[*tx.Hash] = true
+			filteredRecip = append(filteredRecip, tx)
+		}
+	}
+	return sender, filteredRecip
+}
+
+func (c *Cursor) encode() *string {
```
kaladinlight, line 318: I think it makes more sense to return `(string, error)` and just catch the error if unable to encode. Able to remove `logger.Errorf`, `else` and extra var for returning pointer.

go/cmd/cosmos/main.go
asamere: sounds good and simpler

go/coinstacks/cosmos/api/handlers.go
asamere: believe I added this line to try to get at swallowed errors from the issues with the openapi generated client.

go/pkg/api/api.go
asamere: yep, no cursor input equals page 1, no cursor output = last page. open to other suggestions here

go/pkg/cosmos/types.go
asamere: 👍

go/pkg/cosmos/txhistory.go
asamere: thought here is block height is a function of block time, and using unsigned ints where appropriate gives you access to a lot more numbers before you overflow. maybe we should go to BigInt with a custom marshal/unmarshal?

go/pkg/cosmos/txhistory.go
asamere: yep waffled on that, cleaned up some of them before I pushed. I'll clean up the rest 👍

go/pkg/cosmos/txhistory.go
asamere: ❤️  ty

go/pkg/cosmos/txhistory.go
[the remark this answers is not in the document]
asamere: yep left that there for demoing purposes - and to know when to blame the node vs my code for latency

go/pkg/cosmos/txhistory.go
asamere: sounds reasonable

go/pkg/cosmos/txhistory.go
asamere: good 👁️  - I also had the "can't parse just serve page 1" approach we mentioned earlier on

go/pkg/cosmos/txhistory.go
```diff
@@ -126,92 +162,416 @@ func Messages(msgs []sdk.Msg) []Message {
...
+		if len(s.sends) == 0 && !senderDone {
+			_, senderDone, err = s.refill(histTxTypeSender)
+			if err != nil {
+				return nil, errors.Wrap(err, "error re-filling sends")
+			}
+		}
+
+		// get the first tx from each type
+		var topOfSent, topOfRecvd uint64
+		if len(s.sends) > 0 && s.sends[0].Height != nil {
```
kaladinlight, line 414: Were you seeing cases of height null? I didn't think tx_search returned any mempool transactions. And if we do see this case, do we want to just throw it away if not valid, or handle it and show as most recent if mempool? Right now it looks like it could just freeze the set of txs

go/pkg/cosmos/txhistory.go
asamere: This is the entry point, and should move to `GetTxHistory` as your prior comment suggested. Started dev in isolated function and left the skeleton/delegate.

It's initializing the cursor vars here, subtracting one to undo increment before prior page ended. might be able to change the way we increment to remove the need for that. Totally game for a cleaner/functional approach.

kaladinlight:
merged in: https://github.com/shapeshift/unchained/pull/270

kaladinlight:
merged in: https://github.com/shapeshift/unchained/pull/270

PR #260, opened 2022-02-16
refactor: use getParsedTxWithTransfers for clarity

0xApotheosis:
@kaladinlight a quick refactor as an example of what we briefly DMed about.

IMO this makes the logic a little easier to follow and could reduce developer error.

### DIFF
--- a/node/packages/parser/src/ethereum/index.ts
+++ b/node/packages/parser/src/ethereum/index.ts
@@ -56,9 +56,6 @@ export class TransactionParser {
   }
 
   async parse(tx: Tx, address: string, internalTxs?: Array<InternalTx>): Promise<ParseTx> {
-    const sendAddress = tx.vin[0].addresses?.[0] ?? ''
-    const receiveAddress = tx.vout[0].addresses?.[0] ?? ''
-
     // We expect only one Parser to return a result. If multiple do, we take the first and early exit.
     const contractParserResult = await findAsyncSequential<Parser, TxSpecific<ParseTx>>(
       this.parsers,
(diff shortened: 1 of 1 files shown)


node/packages/parser/src/ethereum/index.ts
```diff
@@ -79,6 +76,58 @@ export class TransactionParser {
       value: tx.value,
     }
 
+    const pTxWithTransfers = this.getParsedTxWithTransfers(tx, pTx, address, internalTxs)
```
gomesalexandre, line 79: I know many have OCD about single-letter variables, I think they'd get the same sentiment here 😁 
```suggestion
    const parsedTxWithTransfers = this.getParsedTxWithTransfers(tx, pTx, address, internalTxs)
```

node/packages/parser/src/ethereum/index.ts
```diff
@@ -79,6 +76,58 @@ export class TransactionParser {
...
+  ): Array<Transfer> {
+    if (!new BigNumber(value).gt(0)) return transfers
+
+    const index = transfers?.findIndex((t) => t.type === type && t.caip19 === caip19 && t.from === from && t.to === to)
+    const transfer = transfers?.[index]
+
+    if (transfer) {
+      transfer.totalValue = new BigNumber(transfer.totalValue).plus(value).toString(10)
+      transfer.components.push({ value: value })
+      transfers[index] = transfer
```
gomesalexandre, line 109: I know this is just moved code, but since we return `transfer` and use it like so `        pTx.transfers = this.aggregateTransfer(...args)`, do we actually want to mutate `transfer`?
To make it cleaner and avoid side-effects, we could avoid using references and just (deep) clone the object return a new one with all fields declaratively set (the non-changed ones could be spread from old object)

node/packages/parser/src/ethereum/index.ts
gomesalexandre: Same goes for other `ptx` references that would be `parsedTx`

gomesalexandre (approved):
That looks good to me, a few minor comments.
Not sure how to test it since I'm not familiar with unchained so consider my approval as an "I've checked out the repo locally, went through the logic and made sure it made sense, and ran `yarn build` in `node/packages/parser` and saw nothing broke compared to `develop`)

node/packages/parser/src/ethereum/index.ts
0xApotheosis: 😅 I hate them too - I'll get them fixed up.

0xApotheosis: I'm actually going to leave that out of this PR, as the "pTx' convention is currently used widely, and would touch more files than I planned to in this PR.

I'll do a follow-up 👌

kaladinlight: I think I was doing some extra golang dev at the time and had short var names in mind haha. Feel free to update for more clarity in a follow up PR!

node/packages/parser/src/ethereum/index.ts
kaladinlight: This has been brought up before and is a valid concern. I will make a small issue ticket for this

kaladinlight (approved):
We can keep rolling forward with this approach and just continue to evaluate as we go along. 
The one thing that is striking me as a bit weird is that we have the identical transfer object in the transfers array and now the (buy/sell)Tx object. Feels like metadata should be just that and not necessarily a copy of the data. Also looking at web, all we are using that data for is the keying on the asset and the value sold or bought. 

The other thing I think we should keep in mind is some of this logic is not that complicated to gather if we know the use case we are dealing with (trade/approve/lp/etc) and have all of the transfer info available on the parsed transaction.

I know you are hoisting over this conditional logic from web for starters, but the other improvement we can make with buy/sell specifically is this information is really associated with a trade, so it seems like it should just be included with the trade data. And maybe trade could move into metadata to follow the new pattern.

Just some food for thought as we continue iterating.

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Noting here that we've DMed about this, and I agree with where your head is at.

I'll merge this as-is, but expect more PRs in this direction 👌

PR #261, opened 2022-02-17
refactor: reduce redundancy in ethereum parser tests

0xApotheosis:
Reduce redundancy in Ethereum parser tests.


gomesalexandre (approved):
Nice one! That's super clean 🧽

kaladinlight (approved):
LGTM. The only thing that caught my eye was the naming `parsedSomethingTx` when that payload is actually a transfer which is a subset of the parsed tx, but not a huge deal

PR #265, opened 2022-02-18
Cosmos parse std tx

kaladinlight:


node/packages/parser/src/cosmos/types.ts
```diff
@@ -0,0 +1,38 @@
+export type Network = 'mainnet'
+
+interface Value {
+  amount: string
+  denom: string
```
0xApotheosis, line 5: `denomination`?

node/packages/parser/src/cosmos/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      caip2: caip2.toCAIP2({ chain: ChainTypes.Cosmos, network: toNetworkType(this.network) }),
+      confirmations: -1, // TODO: confirmations not tracked by cosmos coinstack
+      status: Status.Confirmed, // no mempool provided by cosmos coinstack currently, and can be inferred from confirmations when added
+      transfers: [],
+      txid: tx.txid,
+      value: tx.value,
+    }
+
+    // messages make best attempt to track where value is transferring to for a variety of tx types
+    // logs provide more specific information if needed as mor complex tx types are added
```
0xApotheosis, line 43:
```suggestion
    // logs provide more specific information if needed as more complex tx types are added
```

node/packages/parser/src/cosmos/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    tx.messages.forEach((msg) => {
```
0xApotheosis, line 44: I'd like to see us replace this mutating pattern, which we use in many places, with a pure one (e.g. `return tx.messages.map...`, with the usage of `aggregateTransfer` refactored to be pure, too) - fine for now though.

0xApotheosis (approved):

node/packages/parser/src/cosmos/types.ts
kaladinlight: Original key name from cosmos-sdk that I am passing back through unchained, going to leave it as.

node/packages/parser/src/cosmos/index.ts
kaladinlight: Ticket created: https://github.com/shapeshift/unchained/issues/268

PR #267, opened 2022-02-21
feat: parse Yearn transactions

0xApotheosis:
_Note: work in progress._

**Context**
By detecting Yearn transactions with a dedicated parser, we are one step closer to being able to add Yearn-specific logic in `unchained` and pass it to `web` to support additional information when rendering.

**Description**
- Adds Yearn parser, `ethereum/yearn.ts`
- Adds method names (`approve`, `deposit`, `withdraw` etc) and parser used metadata to transactions across all parsers (though zrx is missing method data for now)
- Add additional ABIs.- ShapeShift Router and Yearn
- Adds types to new and existing ABIs


node/packages/parser/src/ethereum/index.ts
```diff
@@ -21,15 +22,18 @@ export interface TransactionParserArgs {
   rpcUrl: string
 }
 
-export type Parser = thor.Parser | uniV2.Parser | zrx.Parser
+export interface GenericParser {
```
kaladinlight, line 25: Nice! We can add a generic tx type to support parsers across all chains as well and hoist to `parser/src/index.ts` as well.

node/packages/parser/src/ethereum/index.ts
```diff
@@ -21,15 +22,18 @@ export interface TransactionParserArgs {
...
+  parse: (tx: Tx) => Promise<Partial<TxSpecific<ParseTx>> | undefined>
```
kaladinlight, line 26: Just noticed we are redundantly using `Partial` here and in the zrx parser function since `TxSpecific` is already a partial type

node/packages/parser/src/ethereum/index.ts
```diff
@@ -114,7 +123,7 @@ export class TransactionParser {
     return transfers
   }
 
-  private getStatus(tx: Tx): Status {
+  private static getStatus(tx: Tx): Status {
```
kaladinlight, line 126: I have noticed a couple updates to static, just curious as to your pattern on making a function static. Anything that doesn't reference this?

I would say that in this case, making a private function static doesn't provide us with any benefit that I know know of.

node/packages/parser/package.json
```diff
@@ -17,6 +17,7 @@
     "@shapeshiftoss/caip": "^1.6.1",
     "@shapeshiftoss/thorchain": "^5.1.1",
     "@shapeshiftoss/types": "^1.17.0",
+    "abi-decoder": "^2.4.0",
```
kaladinlight, line 20: Is there a reason we aren't just using `ethers.js` for abi decoding purposes to avoid bringing in another package?

kaladinlight: It would also reduce the need for the extra `ABI.Item` types being spread throughout I would assume since ethers understands vanilla json arrays

node/packages/parser/src/ethereum/utils.ts
```diff
@@ -17,3 +18,6 @@ export const toNetworkType = (network: Network): NetworkTypes => {
       throw new Error('unsupported network')
   }
 }
+
+const itemToFragment = (item: ABI.Item) => ({ ...item, gas: item.gas?.toString() })
```
kaladinlight, line 22: Assuming this is all `abi-decoder` massaging logic?

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
+import { Tx } from '@shapeshiftoss/blockbook'
+import { TxSpecific as ParseTxSpecific, YearnTx } from '../types'
+import shapeShiftRouter from './abi/shapeShiftRouter'
+import yearnVault from './abi/yearnVault'
+import { addABI, decodeMethod } from 'abi-decoder'
+import { GenericParser } from './index'
+
+const ROUTER_CONTRACT = '0x6a1e73f12018D8e5f966ce794aa2921941feB17E'
```
kaladinlight, line 8: Is this the ShapeShift router contract address? Would be worth specifying since there are various contracts being interacted with in the yearn ecosystem

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class Parser implements GenericParser {
+  supportsTransaction(tx: Tx) {
```
kaladinlight, line 11: Did we want to just nest our early return conditions in `parse` like the rest?

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
...
+    const data = tx.ethereumSpecific?.data
+    const firstVoutAddress = tx.vout && tx.vout[0].addresses && tx.vout[0].addresses[0]
```
kaladinlight, line 13: Prefer optional chaining over &&ing, and as far as blockbook is concerned, this is always the receive address to better describe value vs `firstVoutAddress` if wanting to set to variable.

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
...
+    const contractData = data?.slice(0, 2).concat(data?.slice(34, 74)).toLowerCase()
```
kaladinlight, line 14: This looks like `contractAddress` vs `contractData` and is this supposed to be checking for the address paramter against multiple contract functions that happen to be in the same position?

I think I would rather be explicit in our conditions and verify specific arguments in calling functions than some magic position of an address in the input data.

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (contractData === ROUTER_CONTRACT.toLowerCase()) return true
+
+    return false
+  }
+
+  async parse(tx: Tx): Promise<ParseTxSpecific<YearnTx> | undefined> {
+    if (!this.supportsTransaction(tx)) return
+    addABI([...shapeShiftRouter, ...yearnVault])
+
+    const decoded = (tx.ethereumSpecific?.data && decodeMethod(tx.ethereumSpecific.data)) || { name: '', params: [] }
```
kaladinlight, line 26: Ideally we would have checked for the existence of `ethereumSpecific?.data` in our early returns to prevent the need here

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    if (firstVoutAddress === ROUTER_CONTRACT) return true
+    if (contractData === ROUTER_CONTRACT.toLowerCase()) return true
+
+    return false
+  }
+
+  async parse(tx: Tx): Promise<ParseTxSpecific<YearnTx> | undefined> {
+    if (!this.supportsTransaction(tx)) return
+    addABI([...shapeShiftRouter, ...yearnVault])
```
kaladinlight, line 24: How does decoding resolve if there are matching functions in each abi? Not sure if that is the case, but with loading both and decoding, it seems like a scenario we would want to understand. 

This would be another case in my opinion in which by knowing which contract we are interacting with, we could explicitly decode the input data against the specific abi.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Indeed. I've improved this in https://github.com/shapeshift/unchained/pull/267/commits/712a59eb5a2e8061d2a7730f4f4a319b19369a21 by using descriptive constants.

node/packages/parser/package.json
0xApotheosis: I thought there was a reason, but it was not valid. I've stripped out `abi-decoder` entirely in favour of `ethers` https://github.com/shapeshift/unchained/pull/267/commits/5606fc19cb7ae325866e20a8fa0d124f4330d5c0.

node/packages/parser/src/ethereum/utils.ts
0xApotheosis: Yep, but it's all now gone because https://github.com/shapeshift/unchained/pull/267/commits/5606fc19cb7ae325866e20a8fa0d124f4330d5c0 swaps out `abi-decoder` with `ethers` 👌

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Good catch, resolved in https://github.com/shapeshift/unchained/pull/267/commits/5606fc19cb7ae325866e20a8fa0d124f4330d5c0.

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Ah yes, it's a convention I usually follow for methods that don't touch any state of the class instance. The method, therefore, sits on the class, rather than the instance of the class.

In theory, there is a small, but negligible performance improvement, but I just use it as a way to help conceptualise the level at which the method operates.

I'm cool either way, but I _think_ using [static](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static) is more idiomatic TypeScript.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Yeah that's probably neater - addressed in https://github.com/shapeshift/unchained/pull/267/commits/2953d65c9a978276a520a2f9707dcb3c1381583e.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Nice, it slipped my mind that we could use optional chaining on array accessors. Addressed in https://github.com/shapeshift/unchained/pull/267/commits/2953d65c9a978276a520a2f9707dcb3c1381583e.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: I've fixed this up in https://github.com/shapeshift/unchained/pull/267/commits/2953d65c9a978276a520a2f9707dcb3c1381583e by using the `args._spender` on the decoded transaction. I don't like that the `_` indicates this probably shouldn't be directly accessed, but I'm not sure how else to do it.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Addressed in https://github.com/shapeshift/unchained/pull/267/commits/2953d65c9a978276a520a2f9707dcb3c1381583e.

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: You're right, it doesn't handle it well.

In the updated `ethers` implementation, I've added an ABI detection switch so we explicitly specify which one we're using.

https://github.com/shapeshift/unchained/pull/267/commits/2953d65c9a978276a520a2f9707dcb3c1381583e

node/packages/parser/src/ethereum/index.ts
kaladinlight: That works with me, thanks for the breakdown!

node/packages/parser/src/ethereum/__tests__/mockData/yearnApproval.ts
```diff
@@ -0,0 +1,38 @@
...
+    gasUsed: 46643,
+    gasPrice: '96896127986',
+    data:
+      '0x095ea7b30000000000000000000000006a1e73f12018d8e5f966ce794aa2921941feb17e00000000000000000fffffffffffffffffffffffffffffffffffffffffffffff',
+  },
+}
+
+export default {
+  tx: yearnApproval,
+  txMempool: { ...yearnApproval, blockHeight: -1, confirmations: 0 },
```
kaladinlight, line 37: Def less verbose this way. I would be sure to add `fee: 0` on the `txMempool` case since we do look at this data during parsing, and we just need to be aware that we won't have the `coinSpecificData` and a few `ethereumSpecific` values would be inaccurate if we ever ended up using them.

0xApotheosis: Fixed up in https://github.com/shapeshift/unchained/pull/267/commits/fe9dc0e294d881b50e152a56d1312dbb56b73723, thanks!

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Isn't that what `StandardTx` already is, or do you mean something else?

node/packages/parser/src/ethereum/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const SHAPE_SHIFT_ROUTER_CONTRACT = '0x6a1e73f12018D8e5f966ce794aa2921941feB17E'
+export const YEARN_V2_ROUTER_CONTRACT = '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D'
+export const YEARN_LINK_TOKEN_VAULT = '0x671a912C10bba0CFA74Cfc2d6Fba9BA1ed9530B2'
```
theoboldfrazier, line 3: Do we need the specific LINK Vault in constants? Makes it seem like its a special vault.

theoboldfrazier (approved):
Looks good to me. Left some comments, but nothing blocking.

node/packages/parser/src/ethereum/thor.ts
```diff
@@ -67,14 +69,21 @@ export class Parser {
     // We didn't recognise the sigHash - exit
     if (!result) return
 
+    const decoded = this.abiInterface.parseTransaction({ data: txData })
```
kaladinlight, line 72: Something that we can address on a follow up pr, but we can probably clean up how we are doing `decodeFunctionData` vs `parseTransaction`. Don't see why we can't just use `parseTransaction` at this point so we have both method name and args and then just have a condition to return on unsupported methods

node/packages/parser/src/ethereum/types.ts
```diff
@@ -16,3 +16,12 @@ export interface InternalTx {
 }
 
 export type Network = 'mainnet' | 'ropsten'
+
+export interface yearnTokenVault {
```
kaladinlight, line 20: Pascal case for types and interfaces

node/packages/parser/src/ethereum/uniV2.ts
```diff
@@ -117,4 +60,76 @@ export class Parser {
     const initCodeHash = '0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f' // https://github.com/Uniswap/v2-periphery/blob/dda62473e2da448bc9cb8f4514dadda4aeede5f4/contracts/libraries/UniswapV2Library.sol#L24
     return ethers.utils.getCreate2Address(factoryContract, salt, initCodeHash)
   }
+
+  private async getTransfers(tx: Tx): Promise<Transfer[] | undefined> {
+    const data = tx.ethereumSpecific?.data
+    if (!data) return
+    const sendAddress = tx.vin[0].addresses?.[0] ?? ''
+    return await (async () => {
```
kaladinlight, line 68: No reason to wrap the switch in another anon async func.

node/packages/parser/src/ethereum/utils.ts
```diff
@@ -25,11 +27,20 @@ export const txInteractsWithContract = (tx: Tx, contract: string) => {
   return receiveAddress === contract
 }
 
-export const getStandardTx = (tx: ParseTx | undefined) => (tx?.transfers?.length === 1 ? tx.transfers[0] : undefined)
+export const getYearnTokenVaultAddresses = async (): Promise<Array<string> | undefined> => {
+  const yearnTokenVaultResponse = await axios.get<yearnTokenVault[]>('https://api.yearn.finance/v1/chains/1/vaults/all')
```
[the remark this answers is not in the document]
kaladinlight: We do have an ethers jsonrpc provider available if needed. I still want to track down the v2 registry contract and we can just query on chain (assuming that exists and is accurate). I found one, but it didn't look up to date and may have been the v1 registry or something

node/packages/parser/src/ethereum/index.ts
kaladinlight: Right now all of the input payload tx types are blockbook transactions, but this won't always be the case. `GenericParser<T> { parse: (tx: T) }` so we can specific what kind of generic parser based on backend endpoints

node/packages/parser/src/ethereum/yearn.ts
kaladinlight: The `_` doesn't have the same private meanings as far as solidity is concerned. Typically just used to indicate a function parameter in the case of

```
address contract;

someFunc(address _contract) public {
    contract = _contract;
}
```

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getSigHash, getYearnTokenVaultAddresses } from './utils'
+
+export class Parser implements GenericParser {
+  yearnTokenVaultAddresses: Array<string> | undefined
+
+  async parse(tx: Tx): Promise<TxSpecific<YearnTx> | undefined> {
+    const data = tx.ethereumSpecific?.data
+    if (!data) return
+
+    const abiInterface = this.getAbiInterface(data)
```
kaladinlight, line 16: Might as well return early if no abiInterface and then we can reduce the following optional chaining.

node/packages/parser/src/ethereum/utils.ts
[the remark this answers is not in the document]
kaladinlight: Also feel like we should not return undefined here. Either return the successful results or throw

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,65 @@
...
+    const withdrawSigHash = '0x00f714ce'
+
+    return (() => {
+      switch (txSigHash) {
+        case approvalSigHash:
+          return yearnInterface
+        case depositSigHash:
+          return shapeShiftInterface
+        case withdrawSigHash:
+          return yearnInterface
```
kaladinlight, line 59: Could probably make use of fall through cases to make this a little more readable
```suggestion
        case approvalSigHash:
        case withdrawSigHash:
          return yearnInterface
        case depositSigHash:
          return shapeShiftInterface
```

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,65 @@
...
+export class Parser implements GenericParser {
+  yearnTokenVaultAddresses: Array<string> | undefined
+
+  async parse(tx: Tx): Promise<TxSpecific<YearnTx> | undefined> {
+    const data = tx.ethereumSpecific?.data
+    if (!data) return
+
+    const abiInterface = this.getAbiInterface(data)
+    const decoded = abiInterface?.parseTransaction({ data })
+    const spender = decoded?.args._spender
```
kaladinlight, line 18: This would be dependent on the sigHash detected. I would prefer explicit conditional check on each possible function call check.

node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -0,0 +1,65 @@
...
+    const receiveAddress = tx.vout?.[0].addresses?.[0]
+
+    if (!this.yearnTokenVaultAddresses) {
+      this.yearnTokenVaultAddresses = await getYearnTokenVaultAddresses()
+    }
+
+    if (
```
kaladinlight, line 25: Explicit conditions would probably help clean this up a bit too.

node/packages/parser/src/ethereum/yearn.ts
kaladinlight: Depending on how the explicit conditions pan out, this may not be relevant in retrospect.

node/packages/parser/src/ethereum/utils.ts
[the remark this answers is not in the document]
theoboldfrazier: I know we use the v2 registry contract in our shapeshift yearn routing contract. It points to ens name `v2.registry.ychad.eth`

https://github.com/shapeshift/yearn-router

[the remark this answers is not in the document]
kaladinlight: I ran across that, but the contract didn't look right in etherscan, or at least not what I was expecting. I will check it out again to see if I missed something. Thanks!

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Tidied up in https://github.com/shapeshift/unchained/pull/267/commits/5fecfd3a2687cfdb37e4b0d571c38d4bcca7b628

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: The fall through is nice - added in https://github.com/shapeshift/unchained/pull/267/commits/5fecfd3a2687cfdb37e4b0d571c38d4bcca7b628

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Addressed in https://github.com/shapeshift/unchained/pull/267/commits/5fecfd3a2687cfdb37e4b0d571c38d4bcca7b628

0xApotheosis: Addressed in https://github.com/shapeshift/unchained/pull/267/commits/5fecfd3a2687cfdb37e4b0d571c38d4bcca7b628

node/packages/parser/src/ethereum/uniV2.ts
0xApotheosis: Addressed in https://github.com/shapeshift/unchained/pull/267/commits/5fecfd3a2687cfdb37e4b0d571c38d4bcca7b628

node/packages/parser/src/ethereum/yearn.ts
0xApotheosis: Fixed up in https://github.com/shapeshift/unchained/pull/267/commits/6aef3583e6d760d005cc6034a5335760b4b5c51a

kaladinlight (approved):

node/packages/parser/src/ethereum/thor.ts
0xApotheosis: Fixed up in https://github.com/shapeshift/unchained/pull/267/commits/93a84e28422f0cc76c1ebddfbccfac54883b5310

node/packages/parser/src/ethereum/index.ts
0xApotheosis: Addressed in https://github.com/shapeshift/unchained/pull/292

PR #282, opened 2022-02-25
chore: parser type improvements

0xApotheosis:
- `StandardTx` is already a part of `Tx`, so adding it to `SequencedTx` is not necessary
- `parser` can always be defined


kaladinlight (approved):
Good catch

PR #283, opened 2022-02-28
refactor: reduce duplication in test data

0xApotheosis:


gomesalexandre (approved):
LGTM! 
Note: Usually we'd like to use some kind of maker function with `merge()` instead of spreading objects and overriding defaults, but as
1. we do not have `lodash` / `lodash.merge` in neither `dependencies` nor`devDependencies` and it probably doesn't make sense to install for the sake of testing and
2. it is actually not a big deal here since the spreads are done in place for the mocks anyway and not within any programmatic logic of tests

That's IMO a non-issue, but just leaving this here for the next reviewers!

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
```diff
@@ -1,82 +1,53 @@
...
+    gasUsed: null,
+    gasPrice: '24200000000',
+    data:
+      '0x7ff36ab500000000000000000000000000000000000000000016df982add888d233e7e8c000000000000000000000000000000000000000000000000000000000000008000000000000000000000000070336a5a868cf9e0a9d1b47fbe1f521ad6c5c3640000000000000000000000000000000000000000000000000000000060cd0b8f0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc200000000000000000000000095ad61b0a150d79219dcf64e1e6cc01f0b64c4ce',
   },
 }
+
+export default {
+  tx: ethOnly,
+  txMempool: { ...ethOnly, blockHeight: -1, confirmations: 0, fee: 0, ethereumSpecific: undefined },
```
kaladinlight, line 52: It might be worth just making a mempoolData mock object that we can spread against all of the various mocks to ensure they are written over the same and we don't accidentally forget a field.

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
```diff
@@ -1,82 +1,53 @@
...
-      gasPrice: '24200000000',
-      data:
+  ],
+  blockHeight: -1,
+  confirmations: 0,
+  blockTime: 1624048855,
+  value: '100295077000000000',
+  fees: '0',
+  rbf: true,
+  coinSpecificData: {
```
kaladinlight, line 24: It looks like you took the wrong tx payload for the non mempool mock

node/packages/parser/src/ethereum/__tests__/mockData/ethSelfSend.ts
```diff
@@ -1,65 +1,44 @@
...
-    fees: '399000000000000',
-    ethereumSpecific: {
-      status: 1,
-      nonce: 49,
-      gasLimit: 23100,
-      gasUsed: 21000,
-      gasPrice: '19000000000',
-      data: '0x',
-    },
+    ethereumSpecific: { ...ethSelfSend.ethereumSpecific, status: -1 },
```
kaladinlight, line 41: we can also set `gasUsed` to null. In relation to my previous comment this would all be good to be contained in some mempool overwrite object/helper function

node/packages/parser/src/ethereum/__tests__/mockData/foxExit.ts
```diff
@@ -1,27 +1,68 @@
...
+    ...foxExit,
     blockHeight: -1,
     confirmations: 0,
-    blockTime: 1626984381,
-    value: '0',
     fees: '0',
+    blockHash: undefined,
     rbf: true,
+    ethereumSpecific: { ...foxExit.ethereumSpecific, status: -1 },
     coinSpecificData: {
```
kaladinlight, line 66: Seems fair to bring over coinSpecific data if we had it available just in case, another I am not seeing across all of the mock updates

0xApotheosis:
Ty for the feedback @kaladinlight, and good catch when I grabbed the wrong data!

All should be addressed in https://github.com/shapeshift/unchained/pull/283/commits/ad873bbeb4881bc46cc23a1858b0156df78336fc via the `mempoolMock` helper - much cleaner.

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
```diff
@@ -1,82 +1,54 @@
...
-    value: '46224000000000000',
-    fees: '1050000000000000',
-    ethereumSpecific: {
-      status: 1,
-      nonce: 4605,
-      gasLimit: 21000,
-      gasUsed: 21000,
-      gasPrice: '50000000000',
-      data: '0x',
+  coinSpecificData: {
```
kaladinlight, line 34: To be clear, `coinSpecificData` only exists on the mempool transaction from blockbook. We do not use this right now, so I would be ok with just removing it if we want to make things easy, but I wouldn't want it looking like it is part of the confirmed tx payload

node/packages/parser/src/ethereum/__tests__/mockData/mempoolMock.ts
```diff
@@ -0,0 +1,15 @@
+import { Tx } from '@shapeshiftoss/blockbook'
+
+export const mempoolMock = (tx: Tx, tokenTransfers = false) => {
+  const ethereumSpecificOverride = { status: -1, gasUsed: null }
+  const mempoolSpecific = {
+    blockHeight: -1,
+    confirmations: 0,
+    fees: '0',
+    blockHash: undefined,
+    rbf: true,
```
kaladinlight, line 10: This would technically be a dynamic value determined by how the transaction was broadcast and not always true for every mempool tx case. Another value we don't currently look at however.

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
0xApotheosis: Got it, I was not aware of that! I'll remove it today.

node/packages/parser/src/ethereum/__tests__/mockData/mempoolMock.ts
0xApotheosis: Ah I see - cool if I remove this from the data for now, then?

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
kaladinlight: Yeah no worries at all. Lot's of random data coming back in these payloads that is not obvious haha

node/packages/parser/src/ethereum/__tests__/mockData/mempoolMock.ts
kaladinlight: That works with me.

node/packages/parser/src/ethereum/__tests__/mockData/ethOnly.ts
0xApotheosis: Actioned in https://github.com/shapeshift/unchained/pull/283/commits/eb15ccb4880aaf1be529a5cc883f9b2151f44d97

node/packages/parser/src/ethereum/__tests__/mockData/mempoolMock.ts
0xApotheosis: Actioned in https://github.com/shapeshift/unchained/pull/283/commits/eb15ccb4880aaf1be529a5cc883f9b2151f44d97

kaladinlight (approved):
Thank you!

PR #285, opened 2022-02-28
Use default cors middleware

kaladinlight:


cjthompson:
Can we confirm that the default CORS allow origin header will be "*"?

cjthompson (approved):

Issue #286, opened 2022-03-01
Add the contract a transaction interacted with to a transaction's metadata


0xApotheosis:
This will be used to show the contract a transaction interacted with in `web`.

0xApotheosis:
Closing for now, in line with https://github.com/shapeshift/web/issues/1086#issuecomment-1057480963

PR #288, opened 2022-03-01
feat: add "interactedWith" contract to Yearn metadata

0xApotheosis:
@kaladinlight a little POC for you - add the contract address that the Yearn transaction interacted with to our parser metadata.

I'm also getting the feeling that all parsers will use a shared metadata type, rather than their own, but we'll see.

Supports https://github.com/shapeshift/unchained/issues/286 and https://github.com/shapeshift/web/issues/1086.

### DIFF
--- a/node/packages/parser/src/ethereum/yearn.ts
+++ b/node/packages/parser/src/ethereum/yearn.ts
@@ -41,6 +41,11 @@ export class Parser implements GenericParser<BlockbookTx> {
     if (!data) return
 
     const txSigHash = getSigHash(data)
+    const isSupportedSigHash = txSigHash
+      ? [this.approvalSigHash, this.depositSigHash, this.withdrawSigHash].includes(txSigHash)
+      : false
+    if (!isSupportedSigHash) return
+
     const abiInterface = this.getAbiInterface(txSigHash)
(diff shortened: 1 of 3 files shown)


node/packages/parser/src/ethereum/yearn.ts
```diff
@@ -21,6 +21,11 @@ export class Parser implements GenericParser {
     if (!data) return
 
     const txSigHash = getSigHash(data)
+    const isSupportedSigHash = txSigHash
```
kaladinlight, line 24: Does this provide us with anything that an undefined `abiInterface` doesn't? I also noticed we have the `getAbiInterface` switch wrapped in an unneeded anon function while looking.

0xApotheosis: You're right, there is actually no value added there - removed in https://github.com/shapeshift/unchained/pull/288/commits/2e555c474ee86bd86212c34b7f271124a9fc3a86

0xApotheosis:
Closing this for the moment, as we need to give the functional requirements more thought.

https://github.com/shapeshift/web/issues/1086#issuecomment-1057480963

PR #297, opened 2022-03-03
EBS Rightsize

samortaylor:
This PR will create smaller EBS volumes in the shapeshift unchained cluster. Once synced we will ensure all environments use the smaller EBS volumes. Cut storage size of RabbitMQ and MongoDB in half.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,11 +33,11 @@ aliases:
     mongo-cpu-limit: 512m
     mongo-memory-limit: 512Mi
     mongo-replicas: 3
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -788,6 +788,7 @@ workflows:
           name: deploy bitcoin develop (shapeshift)
           organization: SHAPESHIFT
           pulumi-command: up -f --yes
+          daemon-storage-size: 650Gi
```
kaladinlight, line 791: Won't this be written over by the `<< : *bitcoin-dev` since the value is defined there as well?

PR #302, opened 2022-03-04
Pod CPU Requests

samortaylor:
This PR introduces the ability to set CPU requests explicitly instead of assuming a 1:1 request:limit ratio. This will allow us to request fewer CPU resources per service, while allowing the service to burst beyond what CPU we request when it needs to. 

We're not making any changes to public services CPU requests in this PR, just getting the configuration in place so we can manipulate CPU requests one service at a time, iteratively. We are modifying the request:limit ratio in develop, in order to validate these changes.

As a bonus CPU savings I've reduced RabbitMQ CPU Limits from 2-->1, based on monitoring this service does not need to request more than 1 CPU.

### DIFF
--- a/node/coinstacks/common/pulumi/src/api.ts
+++ b/node/coinstacks/common/pulumi/src/api.ts
@@ -10,6 +10,7 @@ import { buildAndPushImage, Config, hasTag, getBaseHash } from './index'
 
 export interface ApiConfig {
   cpuLimit: string
+  cpuRequest: string
   memoryLimit: string
   replicas: number
   autoscaling: { enabled: boolean; maxReplicas: number; cpuThreshold: number }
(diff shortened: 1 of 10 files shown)


kaladinlight:
The one thing I am seeing is the config is treating `cpuRequest` as required whereas the pulumi code is partially treating it as optional and using limits as the request. This should be treated the same throughout the call stack.

node/coinstacks/common/pulumi/src/api.ts
```diff
@@ -255,6 +256,9 @@ export async function deployApi(
               cpu: config.api.cpuLimit,
               memory: config.api.memoryLimit,
             },
+            requests: {
+              cpu: config.api.cpuRequest ?? config.api.cpuLimit,
+            },
```
kaladinlight, line 261: Another syntax for this would be:
```suggestion
            ...(config.api.cpuRequests && {
              requests: {
                cpu: config.api.cpuRequest,
              }
            }),
```
Which makes use of kube default of setting request to limit if not provided

kaladinlight: The other updates for this would be not require `cpuRequest` in `config.ts` check. And updating for the value to be optional in any config objects `cpuRequest?: string`

kaladinlight (approved):

PR #305, opened 2022-03-04
Clean up cosmos parser

kaladinlight:


go/coinstacks/cosmos/api/api.go
```diff
@@ -220,7 +220,12 @@ func (a *API) TxHistory(w http.ResponseWriter, r *http.Request) {
 
 	cursor := r.URL.Query().Get("cursor")
 
-	pageSize, err := strconv.Atoi(r.URL.Query().Get("pageSize"))
+	pageSizeQ := r.URL.Query().Get("pageSize")
+	if pageSizeQ == "" {
+		pageSizeQ = "10"
```
pastaghost, line 225: Is there any specific reason to default to a page size of 10?

pastaghost (approved):

go/coinstacks/cosmos/api/api.go
kaladinlight: Just felt like a reasonable client side default

PR #307, opened 2022-03-06
Swap 'No Indexer' and 'With Indexer' in README

fastwin0:
The titles of the images don't correspond to the images themselves, we need to swap 'No Indexer' and 'With Indexer'.


0xApotheosis (approved):

kaladinlight (approved):
Good catch, thank you!

PR #314, opened 2022-03-09
feat: replace tendermint client with resty

### LINKED ISSUE #308: Use Resty client for tendermint
### Overview

In the Cosmos SDK Unchained implementation, we are currently using a tendermint client generated from the Cosmos SDK Open API Spec. This helped us get off the ground, and provides value in terms of generated models, however it also removes some of our control with regard to request/response handling.

The [Resty](https://pkg.go.dev/github.com/go-resty/resty/v2) http client package makes it easy to interact with remote http services, has built in retry capabilities, and requires no extra build time activities. We're already using resty in the Cosmos implementation. Remove all references to the OpenAPI generated tendermint client in Unchained. We will continue using OpenAPI to generate the Unchained server spec.

### Reference

The generated client as well as its Makefile reside here: https://github.com/shapeshift/unchained/tree/develop/go/pkg/tendermint

The tendermint client to remove is instantiated here: https://github.com/shapeshift/unchained/blob/22f417c79b21ea8a574eb989b56eaf291888b176/go/pkg/cosmos/cosmos.go#L75

and injected to the HTTPClient https://github.com/shapeshift/unchained/blob/22f417c79b21ea8a574eb989b56eaf291888b176/go/pkg/cosmos/cosmos.go#L87

It is then used downstream such as:
https://github.com/shapeshift/unchained/blob/22f417c79b21ea8a574eb989b56eaf291888b176/go/pkg/cosmos/history.go#L38

Here's an example of the Cosmos (LCD) Resty client being used to retrieve account data:
https://github.com/shapeshift/unchained/blob/22f417c79b21ea8a574eb989b56eaf291888b176/go/pkg/cosmos/account.go#L30

Note the definition of the Account type on Line 18. We'll need to create models for the types we are consuming from the Tendermint API. See the existing generated model for TxSearchResponseResult for reference. https://github.com/shapeshift/unchained/blob/22f417c79b21ea8a574eb989b56eaf291888b176/go/pkg/tendermint/client/model_tx_search_response_result.go#L18

### Acceptance Criteria

- The generated tendermint client [here](https://github.com/shapeshift/unchained/tree/develop/go/pkg/tendermint) no longer exists.
- All functionality that leverages tendermint continues to function as intended (particularly Tx History)

### Ownership

- [X] As the sponsor of this bounty I will review the PR

### Bounty Hunters

- [ ] Join the [ShapeShift Discord](https://discord.gg/shapeshift)
- [ ] Check out the [Contribution Guidelines](https://shapeshiftdao.gitbook.io/getting-started/how-to-contribute/bounties#resources-for-new-contributors-and-bounty-hunters)
- [ ] Include an expected completion timeline in the work plan when you apply for this bounty on [Gitcoin](https://gitcoin.co/shapeshift/bounties)!

trungtin:
resolve #308


go/pkg/cosmos/cosmos.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/go-resty/resty/v2"
 	"github.com/pkg/errors"
 	"github.com/shapeshift/unchained/internal/log"
-	tendermintclient "github.com/shapeshift/unchained/pkg/tendermint/client"
+
```
kaladinlight, line 25:
```suggestion
```

go/pkg/cosmos/types.go
```diff
@@ -143,3 +142,52 @@ type Value struct {
...
+}
+
+// GetIndex returns the Index field value if set, zero value otherwise.
+func (o *TxSearchResponseResultTxs) GetIndex() int32 {
+	if o == nil || o.Index == nil {
+		var ret int32
+		return ret
+	}
+	return *o.Index
+}
```
kaladinlight, line 173: we aren't utilizing this so let's just remove
```suggestion
```

go/pkg/cosmos/history.go
```diff
@@ -19,36 +20,54 @@ type TxState struct {
...
+	send       *TxState
+	tendermint *resty.Client
 }
 
-func (h *History) doRequest(txState *TxState) (*client.TxSearchResponse, error) {
+func (h *History) doRequest(txState *TxState) (*TxSearchResponse, error) {
 	for {
-		res, httpRes, err := h.tendermintClient.InfoApi.TxSearch(h.ctx).PerPage(STATIC_PAGE_SIZE).OrderBy("\"desc\"").Query(txState.query).Page(int32(txState.page)).Execute()
-		if err != nil {
+		var res *TxSearchResponse
```
kaladinlight, line 39: The commit suggestion is unfortunately not working for the code block I would like, but we can correctly type the error response so we can leverage the `SetError()` and reduce the need for the status code check and subsequent unmarshal:
```
		var res *TxSearchResponse
		var e *struct {
			Jsonrpc string `json:"jsonrpc"`
			Id      int32  `json:"id"`
			Error   struct {
				Data string `json:"data"`
			} `json:"error"`
		}

		queryParams := map[string]string{
			"per_page": strconv.Itoa(int(STATIC_PAGE_SIZE)),
			"order_by": "\"desc\"",
			"query":    txState.query,
			"page":     strconv.Itoa(txState.page),
		}

		_, err := h.tendermint.R().SetResult(&res).SetError(&e).SetQueryParams(queryParams).Get("/tx_search")
		if err != nil {
			return nil, errors.Wrap(err, "failed to do request")
		}

		if e != nil {
			// error is returned if page is out of range past page 1, mark as no more transactions
			if strings.Contains(e.Error.Data, "page should be within") {
				txState.hasMore = false
				return &TxSearchResponse{}, nil
			}

			return nil, errors.Errorf("failed to do request: %s", e.Error.Data)
		}
```

kaladinlight: I also noticed your use of `fmt.Sprint` and have been using `strconv.Itoa` elsewhere and would like to standardize on that style.

go/pkg/cosmos/types.go
trungtin: this is used in coinstacks/cosmos/api/handlers.go and other places as well

kaladinlight: My bad, you are correct.

kaladinlight (approved):
Just need to sync up with develop, but looks good to go!

Issue #309, opened 2022-03-08
Ethereum coinstack - Transaction history endpoint


elmutt:
### Overview

We need to re-implement our ethereum tx history rest endpoint to accept and return different data. A few different Interface type definitions will need to be updated before re-implementing the function -- see reference below

Endpoint to be re-written lives here:

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/ethereum/api/src/controller.ts#L148

The new function can look similar to what exists and use blockbook to get most of the needed data.

### Reference

Update or add the following method signatures and interface types:

#### Common
https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/common/api/src/index.ts#L60:
`getTxHistory(pubkey: string, cursor?: string, pageSize?: number): Promise<TxHistory>`

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/common/api/src/models.ts#L49
```
export interface Pagination {
  cursor?: string
}
```

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/common/api/src/models.ts#L64
```
 export interface Tx {
  txid: string
  blockHash?: string
  blockHeight?: number
  timestamp?: number
}
```

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/common/api/src/models.ts#L80
```
export interface TxHistory extends Pagination {
  pubkey: string
  txs: Array<Tx>
}
```

#### Ethereum Specific

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/ethereum/api/src/models.ts
```
export interface TokenTransfer extends Token {
  from: string
  to: string
  value: string
}

export interface EthereumTx extends Tx {
  from: string
  to: string
  confirmations: number
  value: string
  fee: string
  gasLimit: string
  gasUsed?: string
  gasPrice: string
  status: number
  inputData?: string
  tokenTransfers?: Array<TokenTransfer>
}

 export interface EthereumTxHistory extends TxHistory {
  txs: Array<EthereumTx>
}
```

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/ethereum/api/src/controller.ts#L148:
`getTxHistory(@Path() pubkey: string, @Query() cursor?: string, @Query() pageSize?: number): Promise<EthereumTxHistory>`

### Acceptance Criteria

- The `getTxHistory` endpoint is updated according to the reference spec above
- Update to use base64 encoded cursor based pagination
[...]

PR #316, opened 2022-03-10
Transaction history endpoint re-implementation

boyin0x:
#309

- [x] Update getTxHistory endpoint according to the reference spec.
- [x] Update to use base64 encoded cursor based pagination.
- [x] Verify working endpoint at /api/v1/account/{pubkey}/txs (I verified it using `docker-compose up api`)
- [x] Make minimal necessary modifications to bitcoin getTxHistory endpoint in order to pass build. 
- [x] Update all corresponding tsoa comment annotations to reflect all changes.

**A few doubts:**
1. Default values for this tx fields: `to`, `gasLimit` and `gasPrice`. (Currently I used "unavailable")
2. Default values for this tokenTransfers fields: `balance` and `contract` . (Currently I used "unavailable")

Let me know if you need anything else.

### DIFF
--- a/node/coinstacks/ethereum/api/src/controller.ts
+++ b/node/coinstacks/ethereum/api/src/controller.ts
@@ -11,7 +11,7 @@ import {
   SendTxBody,
   ValidationError,
 } from '../../../common/api/src' // unable to import models from a module with tsoa
-import { EthereumAPI, EthereumAccount, Token, GasFees, EthereumTxHistory,EthereumTx, TokenTransfer } from './models'
+import { EthereumAPI, EthereumAccount, Token, GasFees, EthereumTxHistory, EthereumTx, TokenTransfer } from './models'
 
 const INDEXER_URL = process.env.INDEXER_URL
 const INDEXER_WS_URL = process.env.INDEXER_WS_URL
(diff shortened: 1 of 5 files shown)


boyin0x:
Hi @kaladinlight, just let me know if there is anything you want me to change.

boyin0x:
Thanks @kaladinlight, saw your answer in discord, looking forward for your feedback!

node/coinstacks/ethereum/api/src/controller.ts
```diff
@@ -147,29 +147,58 @@ export class Ethereum extends Controller implements BaseAPI, EthereumAPI {
...
+          throw new ApiError('Bad Request', 400, 'Invalid cursor, please use the cursor returned in previous page')
+        }
+        page = Number(cursorDecoded)
+      }
+
+      const data = await blockbook.getAddress(pubkey, page, pageSize, undefined, undefined, 'txs')
+
+      const nextPage = page + 1
+
+      const nextCursor =
```
elmutt, line 167: We can utilize the totalPages response from blockbook to determine if there are more pages instead of using the length of transactions returned.

Theres a possible edgecase in which the final page returns a full page size which will then be treated as if there is another page even though it was the final transaction.

node/coinstacks/ethereum/api/src/controller.ts
```diff
@@ -147,29 +147,58 @@ export class Ethereum extends Controller implements BaseAPI, EthereumAPI {
...
             txid: tx.txid,
-            status: tx.confirmations > 0 ? 'confirmed' : 'pending',
-            from: tx.vin[0].addresses?.[0] ?? 'coinbase',
-            to: tx.vout[0].addresses?.[0],
             blockHash: tx.blockHash,
             blockHeight: tx.blockHeight,
-            confirmations: tx.confirmations,
             timestamp: tx.blockTime,
+            status: tx.ethereumSpecific?.status ?? tx.confirmations > 0 ? 1 : -1, // assuming confirmed=1 pending=-1 based on previous implementation and TransactionParser.getStatus()
+            from: tx.vin[0].addresses?.[0] ?? 'coinbase',
```
elmutt, line 182: Lets go with empty string for all of these default values

elmutt:
@boyin0x This looks great. 

A few small comments but once those are fixed we will pull down and test. If everything is working we can then merge.

Thanks for the pr!

node/coinstacks/ethereum/api/src/controller.ts
boyin0x: Here used empty string for all the default values, also for the from field which was 'coinbase' in the previous implementation.

boyin0x:
Thanks @elmutt, just made the requested changes.

elmutt (approved):

kaladinlight (approved):
Looks great, thanks!

PR #317, opened 2022-03-10
[WIP] Transaction History Endpoint

GMSteuart:


elmutt:
@GMSteuart thanks for the pr. We decided to go with a PR from another contributor because it was submitted first and was closer to what we are looking for (https://github.com/shapeshift/unchained/pull/316 if you want to take a look)

Definitely keep you eyes open for future bounties if any of them look interesting

PR #335, opened 2022-03-16
Cosmos parser delegations

elmutt:


node/packages/client/src/cosmos/parser/index.ts
```diff
@@ -42,41 +43,30 @@ export class TransactionParser {
...
+      parsedTx.transfers = aggregateTransfer(
+        parsedTx.transfers,
+        transferType,
+        this.assetId,
+        msg.from ?? '',
+        msg.to ?? '',
+        value.toString(10),
+        undefined,
+        msg.type
+      )
```
0xApotheosis, line 69: Not a blocker by any means and I know this pattern already exists, but I'd love to see this implemented in a more functional way.

I.e. instead of mutating `parsedTx`'s `transfers` and `fees` in the `forEach`, resolve them upfront and use them in `parsedTx`'s creation.

0xApotheosis (commented):
`subtype` feels too non-specific to me, like we aren't sure what we're actually referring to. 

I'm not familiar with the Cosmos SDK, but does `withdraw_delegator_reward` and `send` better map to `action` rather than `subtype`?

That said, implementing that suggestion would make the `Transfer` type too Cosmos-specific, so I then wonder if instead the `Transfer` type should be given a `data` property to represent chain-specific metadata, like we've done with the additional of `data` on transactions (`StandardTx`).

elmutt:
> `subtype` feels too non-specific to me, like we aren't sure what we're actually referring to.
> 
> I'm not familiar with the Cosmos SDK, but does `withdraw_delegator_reward` and `send` better map to `action` rather than `subtype`?
> 
> That said, implementing that suggestion would make the `Transfer` type too Cosmos-specific, so I then wonder if instead the `Transfer` type should be given a `data` property to represent chain-specific metadata, like we've done with the additional of `data` on transactions (`StandardTx`).

I wanted to keep the logic changes minimal and not try to think too far ahead with what we may or may not need. In my experience thats how you can get into trouble with these things. By adding a single named field I believe have all the additional information we need to represent delegate/undelegate/withdraw/etc with minimal code additions. 

Im all for doing a larger refactor and perhaps adding a more complex "metadata" type object in the future if we need it but this seems sufficient for now without introducing much tech debt and not having to guess what we may or may not need in that object

node/packages/client/src/cosmos/parser/__tests__/cosmos.test.ts
```diff
@@ -45,29 +63,217 @@ describe('parseTx', () => {
...
+          ],
+        },
+      ],
+      txid: 'E34AFB3A28198957040073034E16D4A979B403E672859651B41C207538136ABE',
+      value: '39447',
+      data: {
+        parser: 'cosmos',
+        method: 'withdraw_delegator_reward',
+        extras: {
+          from: 'cosmosvaloper1hdrlqvyjfy5sdrseecjrutyws9khtxxaux62l7',
```
gomesalexandre, line 137: Just to be 100% sure, as a user I should never need to interact directly with `cosmosvalcons` and `cosmosvaloper` addresses right?
Currently in web, we have a validation that only allows sending to `cosmos` HRP and I'm wondering if we should support these two also.

node/packages/client/src/cosmos/parser/index.ts
```diff
@@ -33,48 +42,47 @@ export class TransactionParser {
       status: tx.confirmations > 0 ? Status.Confirmed : Status.Pending, // TODO: handle failed case
       transfers: [],
       txid: tx.txid,
-      value: tx.value,
+      value,
+      data: {
+        parser: 'cosmos',
```
gomesalexandre, line 47: I grepped for `parser: ` and we currently use strings for parsers but it would be great to have an enum for these. Happy to add it after this is merged!

PR #378, opened 2022-04-06
remove debug logs and add nil checks

elmutt:
remove debug logs and add nil checks

### DIFF
--- a/go/pkg/cosmos/block.go
+++ b/go/pkg/cosmos/block.go
@@ -41,17 +41,14 @@ func (s *BlockService) WriteBlock(block *Block, latest bool) {
 }
 
 func (s *BlockService) GetBlock(height int) (*Block, error) {
-	logger.Infof("GetBlock %d", height)
 	s.m.RLock()
 	block, ok := s.Blocks[height]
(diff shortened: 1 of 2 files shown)


samortaylor (approved):

go/pkg/cosmos/block.go
```diff
@@ -63,34 +60,22 @@ func (s *BlockService) GetBlock(height int) (*Block, error) {
...
 
 func (c *HTTPClient) GetBlock(height *int) (*Block, error) {
-	logger.Infof("GetBlock() %s", height)
 	var res *BlockResponse
 	var resErr *struct {
 		RPCErrorResponse
 		Message string `json:"message"`
 	}
-	logger.Info("GetBlock() 1")
 	req := c.tendermint.R().SetResult(&res).SetError(&resErr)
```
asamere, line 68: are we intentionally passing a pointer to a pointer here? 🤔

asamere (approved):
🍭

go/pkg/cosmos/block.go
elmutt: no idea

PR #384, opened 2022-04-14
Account & info endpoints make calls asynchronously

elmutt:
Account & info endpoints now make their calls to cosmos nodes asynchronously

### DIFF
--- a/go/coinstacks/cosmos/api/handlers.go
+++ b/go/coinstacks/cosmos/api/handlers.go
@@ -5,13 +5,13 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
-	"sync"
 
 	"github.com/pkg/errors"
 	"github.com/shapeshift/unchained/pkg/api"
 	"github.com/shapeshift/unchained/pkg/cosmos"
 	"github.com/shapeshift/unchained/pkg/websocket"
 	"github.com/tendermint/tendermint/types"
+	"golang.org/x/sync/errgroup"
 )
 
 type Handler struct {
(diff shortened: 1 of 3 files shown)


go/coinstacks/cosmos/api/handlers.go
```diff
@@ -151,43 +186,86 @@ func (h *Handler) GetInfo() (api.Info, error) {
...
-		return nil, errors.Wrapf(err, "failed to parse apr: %s", apr)
-	}
+	var anyError error
 
-	account, err := h.httpClient.GetAccount(pubkey)
-	if err != nil {
-		return nil, err
-	}
+	var wg sync.WaitGroup
+	wg.Add(6)
```
asamere, line 194: so this is gonna exec all 6 requests to the node simultaneously. what happens as we add more? IMO it would be preferable to throttle the number of number of concurrent requesters to a fixed number. i.e. 3 at a time, 4 at a time or whatever and queue beyond that

go/coinstacks/cosmos/api/handlers.go
```diff
@@ -90,23 +91,57 @@ func (h *Handler) StartWebsocket() error {
...
-	bondedTokens, err := h.httpClient.GetBondedTokens()
+	}()
+	wg.Wait()
+	return totalSupply, annualProvisions, communityTax, bondedTokens, anyError
+}
+
+func (h *Handler) GetInfo() (api.Info, error) {
+	totalSupply, annualProvisions, communityTax, bondedTokens, err := h.GetRawAprData()
 	if err != nil {
 		return nil, err
```
asamere, line 146: are we intentionally doing all or nothing? if `GetUnbondings` fails but everything else succeeds should we hold out on the rest of the data?

elmutt: Keeps it simple in my mind. Its gonna be so rare that any of them fail and the rest succeed

go/coinstacks/cosmos/api/handlers.go
elmutt: Gonna redo this with ErrorGroup to address this issue and make error handling easier

go/coinstacks/cosmos/api/handlers.go
```diff
@@ -90,23 +91,45 @@ func (h *Handler) StartWebsocket() error {
...
+	var communityTax string
+	var bondedTokens string
+	g := new(errgroup.Group)
 
-	communityTax, err := h.httpClient.GetCommunityTax()
-	if err != nil {
-		return nil, err
-	}
+	g.Go(func() error {
+		var err error
```
kaladinlight, line 103: maybe just one top level error variable

kaladinlight (approved):

PR #387, opened 2022-04-16
Foxy parser

DaoDev44:
- Add foxy parser for parsing FOXy transactions


node/packages/client/src/ethereum/parser/foxy.ts
```diff
@@ -0,0 +1,69 @@
...
+      (this.claimWithdrawSigHash = this.abiInterface.getSighash('claimWithdraw'))
+  }
+
+  async parse(tx: BlockbookTx): Promise<TxSpecific | undefined> {
+    if (!txInteractsWithContract(tx, FOXY_STAKING_CONTRACT)) return
+    const txData = tx.ethereumSpecific?.data
+    if (!txData) return
+
+    const txSigHash = getSigHash(txData)
+    const abiInterface = this.getAbiInterface(txSigHash)
```
0xApotheosis, line 33: This feels a little misleading - we already have an `abiInterface` on the class. What this is actually doing is checking to see if the `this.abiInterface` supports any of the supported foxy functions.

Can we rename `getAbiInterface` to something like `abiInterfaceSupportsSigHash`, and then early return on line 34 if not?

Then we could use the interface on our class instead of creating a second one.

node/packages/client/src/ethereum/parser/foxy.ts
```diff
@@ -0,0 +1,69 @@
...
+  readonly claimWithdrawSigHash: string
+
+  constructor() {
+    this.abiInterface = new ethers.utils.Interface(FOXY_STAKING_ABI)
+    // Hard coded staking sigHash because there is two different stake methods
+    // in the foxy staking contract. This gives us the ability to parser the correct one.
+    ;(this.stakeSigHash = '0x7acb7757'),
+      (this.unstakeSigHash = this.abiInterface.getSighash('unstake')),
+      (this.instantUnstakeSigHash = this.abiInterface.getSighash('instantUnstake')),
+      (this.claimWithdrawSigHash = this.abiInterface.getSighash('claimWithdraw'))
```
0xApotheosis, line 24: Is there an advantage to using the comma operator in this way, over just doing the below?

```ts
this.stakeSigHash = '0x7acb7757'
this.unstakeSigHash = this.abiInterface.getSighash('unstake')
this.instantUnstakeSigHash = this.abiInterface.getSighash('instantUnstake')
this.claimWithdrawSigHash = this.abiInterface.getSighash('claimWithdraw')
```

node/packages/client/src/ethereum/parser/foxy.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class Parser implements SubParser {
+  abiInterface: ethers.utils.Interface
+
+  readonly stakeSigHash: string
+  readonly unstakeSigHash: string
+  readonly instantUnstakeSigHash: string
+  readonly claimWithdrawSigHash: string
+
+  constructor() {
```
0xApotheosis, line 17: As this class accepts no arguments, do we actually need a constructor?

0xApotheosis (changes requested):
Nice one @DaoDev44, looking good - added some thoughts.

elmutt (approved):

elmutt:
looks good to me

node/packages/client/src/types.ts
```diff
@@ -54,6 +54,7 @@ export enum TxParser {
   UniV2 = 'uniV2',
   ZRX = 'zrx',
   Thor = 'thor',
+  Foxy = 'foxy',
```
elmutt, line 57: not a big deal for this PR but I wonder if we should eventually separate these into TxParser and "EthSubParser" or something. yearn uniswap, zrx feel like sub parsers of eth to me

node/packages/client/src/ethereum/parser/foxy.ts
DaoDev44: Nope, not at all. I believe this was a linter change. Will fix.

node/packages/client/src/ethereum/parser/foxy.ts
DaoDev44: In order to use the initialized `sigHash` properties, I believe you still need a constructor.

node/packages/client/src/types.ts
DaoDev44: Yea that's good call, and I agree. But I think that makes sense to do in a separate PR.

node/packages/client/src/ethereum/parser/foxy.ts
0xApotheosis: This whole section could become:

```ts
  abiInterface = new ethers.utils.Interface(FOXY_STAKING_ABI)
  // Hard coded staking sigHash because there is two different stake methods
  // in the foxy staking contract. This gives us the ability to parse the correct one.
  stakeSigHash = '0x7acb7757'
  unstakeSigHash = this.abiInterface.getSighash('unstake')
  instantUnstakeSigHash = this.abiInterface.getSighash('instantUnstake')
  claimWithdrawSigHash = this.abiInterface.getSighash('claimWithdraw')
```

node/packages/client/src/ethereum/parser/foxy.ts
0xApotheosis: With the current implementation, yes - see https://github.com/shapeshift/unchained/pull/387/files/29de97a59c4f169870023a013d793e8ec000d83a#r852436841 for an alternative to consider.

node/packages/client/src/ethereum/parser/foxy.ts
DaoDev44: Ahh gocha! Makes sense

0xApotheosis: Get that updated and we'll get this bad boy merged!

0xApotheosis (approved):
Nice one.

Issue #396, opened 2022-04-27
Flatten exports from CAIP packages


0xApotheosis:
### Overview

Use flattened exports from the outcome of https://github.com/shapeshift/lib/issues/548

### References and additional details

Like https://github.com/shapeshift/web/pull/1591, but for `unchained`.

### Acceptance Criteria

Functions like `toCAIP19` or `fromCAIP19` are imported directly from the CAIP packages.

kaladinlight:
Now tracked in lib: https://github.com/shapeshift/lib/issues/619

PR #397, opened 2022-04-27
refactor(caip): flatten CAIP imports

0xApotheosis:
This imports CAIP utils from the new flattened exports.

⚠️ https://github.com/shapeshift/lib/pull/560 should be merged and published first - it contains the flattened exports. 

Once that's done, we should run `yarn` one more time here to update the yarn lockfile.

Closes https://github.com/shapeshift/unchained/issues/396
Partially completes https://github.com/shapeshift/lib/issues/548

### DIFF
--- a/node/packages/client/package.json
+++ b/node/packages/client/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "@shapeshiftoss/blockbook": "^6.9.0",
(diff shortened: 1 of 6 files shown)


kaladinlight:
unchained client is being moved into lib. these changes will need to be brought over upon introduction there

kaladinlight:
package officially exists in lib. changes will need to be implemented there.

PR #408, opened 2022-05-03
Add connection timeout

kaladinlight:
- Throw if connection does not succeed within 5 seconds


0xdef1cafe (approved):
that'll do the trick

PR #417, opened 2022-05-11
osmosis integration

elmutt:
osmosis integration


go/coinstacks/osmosis/api/handlers.go
```diff
@@ -155,7 +172,7 @@ func (h *Handler) GetTxHistory(pubkey string, cursor string, pageSize int) (api.
 			GasUsed:       t.TendermintTx.TxResult.GasUsed,
 			Index:         int(t.TendermintTx.GetIndex()),
 			Memo:          t.SigningTx.GetMemo(),
-			Messages:      osmosis.Messages(t.CosmosTx.GetMsgs()),
+			Messages:      cosmos.Messages(t.CosmosTx.GetMsgs()),
```
gomesalexandre, line 175: Is this named `cosmos` because we're referring to Cosmos SDK vs. osmosis the chain?

gomesalexandre (commented):
Given my knowledge of Go and the go unchained coinstacks, this was mostly about taking a look at the logic to understand how the integration works on the unchained side and double-check there's nothing weird in the diff on the logic side vs. proper reviewing the logic, but from looking at the structs/types/swagger docs routes, this looks good.

go/cmd/osmosis/main.go
```diff
@@ -47,23 +46,31 @@ func main() {
 		}
 	}
 
-	encoding := cosmos.NewEncoding(gammtypes.RegisterInterfaces, lockuptypes.RegisterInterfaces)
+	encoding := cosmos.NewEncoding()
```
kaladinlight, line 49: We will want to register these osmosis specific types for tx decoding

go/coinstacks/osmosis/api/handlers.go
kaladinlight: Yes, this and the Messages call in `handleTx` should be using `osmosis.Messages`

go/coinstacks/osmosis/api/handlers.go
```diff
@@ -174,6 +191,150 @@ func (h *Handler) GetTxHistory(pubkey string, cursor string, pageSize int) (api.
...
+
+	apr, _, err := new(big.Float).Parse(info.(Info).APR, 10)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse apr: %s", apr)
+	}
+
+	return h.httpClient.GetValidator(address, apr)
+}
+
+func (h *Handler) getAPRData() (*APRData, error) {
```
kaladinlight, line 230: Do any of these calls actually work? I would prefer we don't include calls or data that we aren't getting yet or that may not work on osmosis. We can add the necessary data in a follow up PR.

go/coinstacks/osmosis/api/handlers.go
```diff
@@ -174,6 +191,150 @@ func (h *Handler) GetTxHistory(pubkey string, cursor string, pageSize int) (api.
...
+
+		return err
+	})
+
+	err := g.Wait()
+
+	return aprData, err
+}
+
+func (h *Handler) getAccountData(pubkey string, denom string, apr *big.Float) (*AccountData, error) {
```
kaladinlight, line 296: Can you confirm that the account data is returned correctly as it stands?

kaladinlight (commented):
for whatever reason, thorchain main.go is using the osmosis.api which changed the input params causing build to fail. Can you update to use the thorchain api.New?

go/coinstacks/osmosis/api/handlers.go
elmutt: removed

go/cmd/osmosis/main.go
elmutt: done

go/coinstacks/osmosis/api/handlers.go
elmutt: works in swagger docs. seems to be right at first glance:

```

  "balance": "16628631",
  "unconfirmedBalance": "0",
  "pubkey": "osmo1fx4jwv3aalxqwmrpymn34l582lnehr3eg40jnt",
  "accountNumber": 120668,
  "sequence": 112,
  "assets": [
    {
      "amount": "10329755410319112270",
      "denom": "gamm/pool/1"
    },
    {
      "amount": "723781",
      "denom": "ibc/27394FB092D2ECCD56123C74F36E4C1F926001CEADA9CA97EA622B25F41E5EB2"
    },
    {
      "amount": "122009731",
      "denom": "ibc/9712DBB13B9631EDFA9BF61B55F1B2D290B2ADB67E3A4EB3A875F3B6081B3B84"
    },
    {
      "amount": "1191860",
      "denom": "ibc/BE1BB42D4BE3C30D50B68D7C41DB4DFCE9678E8EF8C539F6E6A9345048894FCC"
    }
  ],
  "delegations": [
    {
      "validator": {
        "address": "osmovaloper1pjmngrwcsatsuyy8m3qrunaun67sr9x74vvvdk",
        "moniker": "Cypher Core",
        "jailed": false,
        
        ```

File: go/coinstacks/osmosis/api/handlers.go [hunk 5]
```diff
@@ -88,39 +92,54 @@ func (h *Handler) StartWebsocket() error {
...
+// TODO add osmosis apr
 func (h *Handler) GetInfo() (api.Info, error) {
 	info := Info{
 		BaseInfo: api.BaseInfo{
 			Network: "mainnet",
 		},
+		TotalSupply:      "0",
+		BondedTokens:     "0",
+		AnnualProvisions: "0",
+		CommunityTax:     "0",
```
kaladinlight, line 104: We can remove these values and from the type until we officially add proper apr calculations. Leaving `APR` since that is a required value to be passed into `getAccountData`

go/coinstacks/osmosis/api/types.go
```diff
@@ -63,3 +88,23 @@ type TxHistory struct {
...
+type AccountData struct {
+	Account       *cosmos.Account
+	Balance       *cosmos.Balance
+	Delegations   []cosmos.Delegation
+	Redelegations []cosmos.Redelegation
+	Unbondings    []cosmos.Unbonding
+	Rewards       []cosmos.Reward
+}
+
+type APRData struct {
```
kaladinlight, line 101: Remove until we officially add osmosis apr

kaladinlight (approved):

PR #458, opened 2022-07-18
fix doge xpub api param detection

asamere:


0xdef1cafe (approved):

kaladinlight (approved):

gomesalexandre (approved):
Looks dgub to me (untested, but from Googling out Dogecoin xpubs this is correct)

PR #476, opened 2022-08-11
feat: broadcast osmo txs to keplr

elmutt:
Keplr allows broadcasting with 0 fees. we want to do this for osmo swapper to work even if the user has no balance


toshiSat (approved):

kaladinlight (approved):
This works, though we will really want to circle back around and pull out the osmosis specific logic from the base cosmossdk package and bring it into the osmosis coinstack, so we don't continue to pollute the base cosmossdk logic

PR #478, opened 2022-08-16
fix: get addrs from events instead of only messages

elmutt:
Get relevant websocket registry addresses from events and messages insteas of just events.

Needed for some transactions (ibc receive) that do not have any messages

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,7 @@ aliases:
     rpc-url: https://osmosis-1--rpc--archive.datahub.figment.io
     grpc-url: https://osmosis-1--grpc--archive.datahub.figment.io
     ws-url: wss://osmosis-1--rpc--archive.datahub.figment.io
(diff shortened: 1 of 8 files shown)


kaladinlight:
Commenting here to reflect dms.

- Pull event log address detection logic into common `cosmos/tx.go` for reuse in thorchain/osmosis
- Prefer early continue/return to reduce indentation depth
- Create issue to handle `/ibc.core.channel.v1.MsgRecvPacket` message type

elmutt:
https://github.com/shapeshift/unchained/issues/480

> Commenting here to reflect dms.
> 
> * Pull event log address detection logic into common `cosmos/tx.go` for reuse in thorchain/osmosis
> * Prefer early continue/return to reduce indentation depth
> * Create issue to handle `/ibc.core.channel.v1.MsgRecvPacket` message type

https://github.com/shapeshift/unchained/issues/480

kaladinlight (approved):

PR #489, opened 2022-08-30
fix: urls and resty client creation

kaladinlight:


gomesalexandre (approved):
Semantically looks good, but take my review as a surface review as this is untested, go, and unchained

PR #493, opened 2022-08-31
use events to further build out message payloads

kaladinlight:


gomesalexandre (commented):
Conceptual but not effective stamp as this is unchained go coinstack and I'm not familiar with both the language and the logic of this, but conceptually, and looking at the Message/Events from the [`bank`](https://docs.cosmos.network/v0.45/modules/bank/04_events.html#msgsend) and [`distribution`](https://docs.cosmos.network/v0.45/modules/distribution/06_events.html#msgwithdrawdelegatorreward) modules as well as the shape of the [`MsgSend`](https://docs.cosmos.network/v0.45/modules/bank/03_messages.html#msgsend) message and the [`MsgRecvPacket`](https://pkg.go.dev/github.com/cosmos/cosmos-sdk/x/ibc/core/04-channel/types#MsgRecvPacket) type from cosmos-sdk, this looks good to me

PR #508, opened 2022-09-20
fix: revert back to using self hosted nodes

kaladinlight:


0xean (approved):

gomesalexandre (approved):
Compared to fb014b6cad873d3b7ad6f6f87dea37f8e44fbeec which is when we started switching to Infura, hunks reverted LGTM 🥮

PR #524, opened 2022-10-05
fix: use upgrade header to check for websocket connection

kaladinlight:
for unknown reasons, websocket connection are not handled correctly for taxistake endpoints via cloudflare dns. this is an attempt to check for websocket upgrades in a better way and hopefully clear up the proxy issue :crossed_fingers:


0xdef1cafe:
🤞

PR #533, opened 2022-10-21
feat: validator tx history

lmyslinski:
Added endpoints and logic for fetching the transaction history of a given validator address for cosmos-based chains


go/coinstacks/cosmos/api/api.go
```diff
@@ -45,7 +45,7 @@ type API struct {
 }
 
 func New(httpClient *cosmos.HTTPClient, grpcClient *cosmos.GRPCClient, wsClient *cosmos.WSClient, blockService *cosmos.BlockService, swaggerPath string) *API {
-	r := mux.NewRouter()
+	router := mux.NewRouter()
```
lmyslinski, line 48: I'll rename the one-letter vars as I go along if you don't mind. Concise is good but one letter seems a bit too far

kaladinlight: this is a golang naming convention. you can see example docs: https://github.com/gorilla/mux#examples

the idea is that variables close to where they are used and where they are just struct references can be labeled very concisely. same reason handler function convention is single letter as well (ex. `func (h *Handler) StartWebsocket() error`)

go/pkg/api/api.go
```diff
@@ -150,9 +150,9 @@ type TxBody struct {
 	RawTx string `json:"rawTx"`
 }
 
-// swagger:parameters GetAccount GetValidator
+// swagger:parameters GetAccount
 type PubkeyParam struct {
-	// Account address or xpub
+	// Account public key
```
kaladinlight, line 155: throughout the stack we use the pubkey terminology to indicate either an address or xpub as they are both public identifiers. would leave this be and just use the `PubkeyParam` and `PaginatedPubkeyParam` structs for validator endpoint as well

go/pkg/api/api.go
```diff
@@ -169,6 +169,25 @@ type PaginatedPubkeyParam struct {
 	PageSize int `json:"pageSize"`
 }
 
+// swagger:parameters GetValidator
+type ValidatorAddrParam struct {
```
kaladinlight, line 173: Remove in favor of `PubkeyParam`

go/coinstacks/cosmos/api/api.go
```diff
@@ -81,47 +81,47 @@ func New(httpClient *cosmos.HTTPClient, grpcClient *cosmos.GRPCClient, wsClient
...
 
 	v1Transaction := v1.PathPrefix("/tx").Subrouter()
-	v1Transaction.HandleFunc("/{txid}", a.Tx).Methods("GET")
+	v1Transaction.HandleFunc("/{txid}", apiRef.Tx).Methods("GET")
 
 	v1Gas := v1.PathPrefix("/gas").Subrouter()
-	v1Gas.HandleFunc("/estimate", a.EstimateGas).Methods("POST")
+	v1Gas.HandleFunc("/estimate", apiRef.EstimateGas).Methods("POST")
 
 	v1Validators := v1.PathPrefix("/validators").Subrouter()
```
kaladinlight, line 114: we can either break this out into two subrouters for `/validators` and `/validator` to allow us to use the middleware on the `/validator` subrouter, or if we want to keep only a single subrouter, we will need to add the `cosmos.ValidateValidatorPubkey` to any handlers with `{pubkey}` path param

go/coinstacks/cosmos/api/api.go
lmyslinski: Makes sense 👍

go/pkg/cosmos/history.go
```diff
@@ -49,6 +49,27 @@ func NewDefaultSources(client *HTTPClient, pubkey string, formatTx func(*coretyp
...
+
+		return txs, nil
+	}
+
+	return map[string]*TxState{
+		"send":    NewTxState(true, fmt.Sprintf(`"delegate.validator='%s'"`, pubkey), request),
+		"receive": NewTxState(true, fmt.Sprintf(`"unbond.validator='%s'"`, pubkey), request),
+	}
+}
+
```
lmyslinski, line 72: Ideally the messages to parse would be passed in from the outside but I don't want to mess with data flow too much at this point

lmyslinski: @kaladinlight Btw how did you establish the messages to use here? I've checked all of Tendermind RPC api docs as well as Figment ones but I didn't find any detailed overview of the data model used in responses

go/pkg/cosmos/api.go
```diff
@@ -148,31 +172,33 @@ func (a *API) Account(w http.ResponseWriter, r *http.Request) {
...
+// swagger:route GET /api/v1/validators/{pubkey}/txs v1 ValidatorTxHistory
+//
+// Get paginated transaction history for a validator.
+//
+// responses:
+//
+//	200: TxHistory
+//	400: BadRequestError
+//	500: InternalServerError
+func (a *API) ValidatorTxHistory(w http.ResponseWriter, r *http.Request) {
```
kaladinlight, line 195: it's not the most clear, but I have been adding the validator functions to `coinstacks/cosmos/api/api.go` and `coinstacks/osmosis/api/api.go` explicitly at this time. The reason being, all cosmos-sdk coinstacks extend this base cosmos API, but not all of them support the staking module (ex. thorchain at this time). With this being declared here, the thorchain api will pick up the swagger annotations and show it as an available route that doesn't actually exist.

Until this can be composed more intelligently, can you please move this into the cosmos and osmosis coinstack api.go files directly?

go/pkg/cosmos/handler.go
```diff
@@ -28,6 +28,7 @@ type RouteHandler interface {
 	GetInfo() (api.Info, error)
 	GetAccount(pubkey string) (api.Account, error)
 	GetTxHistory(pubkey string, cursor string, pageSize int) (api.TxHistory, error)
+	GetValidatorTxHistory(pubkey string, cursor string, pageSize int) (api.TxHistory, error)
```
kaladinlight, line 31: In the same vein, remove this function from the interface as it should align with the functions declared in `pkg/cosmos/api.go`. It would probably make more sense for this interface to be declared in there as well so it can be made more clear what the interface is directly associated with.
```suggestion
```

go/pkg/cosmos/history.go
kaladinlight: When you say messages, are you referring to the query string param associated with the tx event (ie. `delegate.validator`)?

go/pkg/cosmos/history.go
```diff
@@ -49,6 +49,27 @@ func NewDefaultSources(client *HTTPClient, pubkey string, formatTx func(*coretyp
 	}
 }
 
+func ValidatorHistorySources(client *HTTPClient, pubkey string, formatTx func(*coretypes.ResultTx) (*Tx, error)) map[string]*TxState {
```
kaladinlight, line 52:
```suggestion
func ValidatorTxHistorySources(client *HTTPClient, pubkey string, formatTx func(*coretypes.ResultTx) (*Tx, error)) map[string]*TxState {
```

go/pkg/cosmos/history.go
```diff
@@ -49,6 +49,27 @@ func NewDefaultSources(client *HTTPClient, pubkey string, formatTx func(*coretyp
...
+	request := func(query string, page int, pageSize int) ([]HistoryTx, error) {
```
kaladinlight, line 53: dupe request function logic if you want to DRY up at all

go/pkg/cosmos/history.go
```diff
@@ -49,6 +49,27 @@ func NewDefaultSources(client *HTTPClient, pubkey string, formatTx func(*coretyp
...
+		for _, tx := range result.Txs {
+			txs = append(txs, &ResultTx{ResultTx: tx, formatTx: formatTx})
+		}
+
+		return txs, nil
+	}
+
+	return map[string]*TxState{
+		"send":    NewTxState(true, fmt.Sprintf(`"delegate.validator='%s'"`, pubkey), request),
+		"receive": NewTxState(true, fmt.Sprintf(`"unbond.validator='%s'"`, pubkey), request),
```
kaladinlight, line 69:
```suggestion
		"delegate":    NewTxState(true, fmt.Sprintf(`"delegate.validator='%s'"`, pubkey), request),
		"undelegate": NewTxState(true, fmt.Sprintf(`"unbond.validator='%s'"`, pubkey), request),
```

go/pkg/cosmos/history.go
kaladinlight: These events are not really documented to my knowledge. I either just look at an example transaction and inspect the `logs[].events[]` payload for a specific value needed as the search query:

image

You can see type `delegate` and key `validator`. All events are indexed by default using `type.key` as lookup (ie. `delegate.validator`).

Alternatively, you can view the source code directly. Each module will declare events: https://github.com/cosmos/cosmos-sdk/blob/ff4f995b185f0b7e318bd0a47d0b1aa824406d8b/x/staking/types/events.go#L9

and you can then look for where the actual event is being created with the associated key/val pairs: https://github.com/cosmos/cosmos-sdk/blob/ff4f995b185f0b7e318bd0a47d0b1aa824406d8b/x/staking/keeper/msg_server.go#L242

go/pkg/cosmos/api.go
lmyslinski: Makes sense, I moved the endpoints

go/pkg/cosmos/history.go
lmyslinski: I give up on this one. I've tried rewriting `TxHistorySources` in a few different ways but they all lead to nowhere but it's super hard to get the big picture when everything is called a `request` and defined in an anonymous nested lambda. I'm strongly against using anonymous fn's for anything but the simplest operations. I wanted to unwrap this function into separate bits and couldn't really get there

lmyslinski: Imo this really needs a more comprehensive rewrite, since the logic is split all over the place

kaladinlight: Yeah, one of the most complicated parts of the codebase thus far. I am down to brainstorm some ideas with you or will do a follow up to try and address the situation a bit better.

go/coinstacks/cosmos/api/api.go
```diff
@@ -166,3 +169,27 @@ func (a *API) GetValidator(w http.ResponseWriter, r *http.Request) {
...
+//	200: TxHistory
+//	400: BadRequestError
+//	500: InternalServerError
+func (a *API) ValidatorTxHistory(w http.ResponseWriter, r *http.Request) {
+	validatorAddr := mux.Vars(r)["pubkey"]
+	cursor, pageSize := a.ValidatePagingParams(w, r)
+	println(validatorAddr, cursor, pageSize)
+
+	txHistory, err := a.handler.GetValidatorTxHistory(validatorAddr, cursor, pageSize)
+
```
kaladinlight, line 188:
```suggestion
```

go/coinstacks/osmosis/api/api.go
```diff
@@ -168,3 +168,27 @@ func (a *API) GetValidator(w http.ResponseWriter, r *http.Request) {
...
+//
+// responses:
+//
+//	200: TxHistory
+//	400: BadRequestError
+//	500: InternalServerError
+func (a *API) ValidatorTxHistory(w http.ResponseWriter, r *http.Request) {
+	validatorAddr := mux.Vars(r)["pubkey"]
+	cursor, pageSize := a.ValidatePagingParams(w, r)
+	println(validatorAddr, cursor, pageSize)
```
kaladinlight, line 184:
```suggestion
```

go/coinstacks/osmosis/api/api.go
```diff
@@ -168,3 +168,27 @@ func (a *API) GetValidator(w http.ResponseWriter, r *http.Request) {
...
+
+	txHistory, err := a.handler.GetValidatorTxHistory(validatorAddr, cursor, pageSize)
+
```
kaladinlight, line 187:
```suggestion
```

kaladinlight (approved):

go/pkg/cosmos/history.go
lmyslinski: sounds great, let's give it a shot

PR #534, opened 2022-10-24
feat: websocket detection of endBlockEvents

kaladinlight:
- add `EndBlockEventHandler` to allow processing end block events over websockets to synthesize transactions like thorchain swaps
- shared `getTxFromEndBlockEvents` function for use by websocket handler and tx history
- add extra data to `ResultTx` that we have available to improve performance
- add `eventCache` to `getTxFromEndBlockEvents` so we don't have to have to `ParseBlockEvents` on each end block event in the block.
- `getTxFromEndBlockEvents` was updated to take an event index so we can more easily publish a single synthesized transaction at a time over websockets instead of making that interface more complicated to handle arrays of txs with associated addresses.

image
image


gomesalexandre (approved):
LGTM, obviously take this stamp as a little stamp since this is golang and my understanding of it is very much limited to what my brains' lexer can do of these tokens, but from reading this through twice and pairing / being walked through the domain specific logic of what this solves, and how this solves it, this looks sane

PR #541, opened 2022-10-28
chore: upgrade thorchain node 1.98.1

0xdef1cafe:


.circleci/config.yml
```diff
@@ -1588,4 +1588,4 @@ workflows:
           requires:
             - approve thorchain coinstack (taxistake)
           <<: *thorchain
-          <<: *only-main
\ No newline at end of file
+          <<: *only-main
```
gomesalexandre, line 1591: revert EOF?

gomesalexandre (approved):
1 char change looking sane, question about the other

.circleci/config.yml
kaladinlight: not a huge deal, let's get it in so we can get thorchain back online

PR #561, opened 2022-11-29
fix: avax config and host match

kaladinlight:
- open host traffic
- use ephemeral certs for unique NodeID on shared volume snapshots
- fix incorrect port mapping
- allow both prod and dev dns to be directed to either cluster


0xApotheosis (approved):
Untested, but this looks sane and I trust big kev.

PR #562, opened 2022-11-30
fix: mutex lock to prevent concurrent write panic

kaladinlight:
- gorilla websockets do not support concurrent writes. add mutex lock to prevent this panic case

### DIFF
--- a/go/pkg/websocket/websocket.go
+++ b/go/pkg/websocket/websocket.go
@@ -3,6 +3,7 @@ package websocket
 import (
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


go/pkg/websocket/websocket.go
```diff
@@ -133,11 +135,17 @@ func (c *Connection) Stop() {
...
 	c.ticker.Stop()
-	_ = c.conn.WriteMessage(websocket.CloseMessage, []byte{})
+	_ = c.send(websocket.CloseMessage, []byte{})
 	c.conn.Close()
 	close(c.msgChan)
 }
 
+func (c *Connection) send(messageType int, data []byte) error {
+	c.m.Lock()
+	defer c.m.Unlock()
```
0xdef1cafe, line 145: not a golang wizard but i assume this defer unlock will unlock the mutex when we exit the scope of this fn?

kaladinlight: correct

PR #563, opened 2022-11-30
feat: handle osmosis swap_exact_amount_in

kaladinlight:
- don't create transfer messages for failed transactions (no value transferred)
- create a message for each side of the swap_exact_amount_in event. (log parsing currently does not support handling of multiple values for same attribute key resulting the `NOTE:` logic)


elmutt (approved):

gomesalexandre (approved):
Small stamp, untested but conceptually LGTM from looking at Cosmos SDK types go package

Issue #469, opened 2022-08-02
Add prefix to bitcoincash cashaddr addresses that don't include `bitcoincash:` prefix


kaladinlight:
### Overview

Blockbook doesn't understand cashaddr addresses without the `bitcoincash:` prefix, add if not provided as part of the address

### References and additional details

https://github.com/shapeshift/unchained/blob/develop/node/coinstacks/bitcoincash/api/src/controller.ts

### Acceptance Criteria

- prefix `q` addresses with `bitcoincash`

PR #574, opened 2022-12-06
feat: prefix q cashaddr with bitcoincash:

gomesalexandre:
#### Description

Untested with WS, tested with API endpoints which use the same underlying implementation so we should be 👜.

This PR makes sure we always prefix bitcoincash addresses. Current implementation assumes `bc` or `bitcoincash` prefixes, and simply returns the unparsed address if none of these two prefixes are present, however, blockbook actually requires a prefix.

While at it, also removes the BIP32 / URL Scheme prefix checks for addresses in `formatAddress`. `formatAddress` should take an address as input, if we pass an URL scheme in a method supposed to parse addresses, we have bigger problems. 

#### Issue

- closes https://github.com/shapeshift/unchained/issues/469

##### This diff

image

#### Prod endpoint

image

### DIFF
--- a/node/coinstacks/avalanche/api/src/app.ts
+++ b/node/coinstacks/avalanche/api/src/app.ts
@@ -8,7 +8,6 @@ import {
   middleware,
   ConnectionHandler,
   Registry,
-  AddressFormatter,
   BlockHandler,
   TransactionHandler,
 } from '@shapeshiftoss/common-api'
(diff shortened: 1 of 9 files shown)


kaladinlight:
- Looking at this now, the wrapped `addressFormatter` is pretty redundant and should be removed. All of the `formatAddress` functions are already an `AddressFormatter`, mind cleaning this up?
- `getTxHistory` will need this same format logic on address type call
- might as well update for other coinstacks to keep pattern consistent (btc, ltc)

PR #606, opened 2023-01-29
feat(osmosis): parse join_pool, exit_pool messages

pastaghost:
* adds support for osmosis LP messages


go/coinstacks/osmosis/tx.go
```diff
@@ -29,13 +29,64 @@ func ParseMessages(msgs []sdk.Msg, events cosmos.EventsByMsgIndex) []cosmos.Mess
 			}
 			messages = append(messages, message)
 		case *gammtypes.MsgJoinPool:
-			message := cosmos.Message{
-				Addresses: []string{v.Sender},
-				From:      v.Sender,
-				Type:      v.Type(),
-				Value:     cosmos.CoinToValue(&v.TokenInMaxs[0]),
+			tokenInValues := cosmos.PoolEventAmountToValues(events["0"]["pool_joined"]["tokens_in"])
```
kaladinlight, line 32: we can leverage existing functionality over new coin parsers:
```
tokenInCoins, err := sdk.ParseCoinsNormalized(events["0"]["pool_joined"]["tokens_in"])
if err != nil {
	fmt.Println(err)
}

// token in 0
Value: cosmos.CoinToValue(&tokenInCoins[0])

// token in 1
Value: cosmos.CoinToValue(&tokenInCoins[1])
```

go/coinstacks/osmosis/tx.go
```diff
@@ -29,13 +29,64 @@ func ParseMessages(msgs []sdk.Msg, events cosmos.EventsByMsgIndex) []cosmos.Mess
...
+					Value:     tokenInValues[1],
+				},
+				// token out (lp token)
+				{
+					Addresses: []string{v.Sender},
+					Origin:    v.Sender,
+					From:      events["0"]["transfer"]["sender"],
+					To:        v.Sender,
+					Type:      v.Type(),
+					Value:     cosmos.SerializedPoolTransferStringtoValue(events["0"]["transfer"]["amount"]),
```
kaladinlight, line 57: also prefer using message index from for loop to handle possible multiple join/exit messages in a single transaction
```suggestion
					Value:     cosmos.SerializedPoolTransferStringtoValue(events[strconv.Itoa(i)]["transfer"]["amount"]),
```

go/coinstacks/osmosis/tx.go
```diff
@@ -29,13 +29,64 @@ func ParseMessages(msgs []sdk.Msg, events cosmos.EventsByMsgIndex) []cosmos.Mess
...
-				From:      v.Sender,
-				Type:      v.Type(),
-				Value:     cosmos.CoinToValue(&v.TokenInMaxs[0]),
+			tokenInValues := cosmos.PoolEventAmountToValues(events["0"]["pool_joined"]["tokens_in"])
+			msgs := []cosmos.Message{
+				// token in 0
+				{
+					Addresses: []string{v.Sender},
+					Origin:    v.Sender,
+					From:      v.Sender,
```
kaladinlight, line 38: ideally we would want the `To` address, however there is an issue with event parsing in which keys with the same name overwrite each other... I wouldn't think this would be a blocker at this time, but the data is available if we want to get it, will just require some work with the events data structure

go/coinstacks/osmosis/tx.go
```diff
@@ -29,13 +29,64 @@ func ParseMessages(msgs []sdk.Msg, events cosmos.EventsByMsgIndex) []cosmos.Mess
...
+					Origin:    v.Sender,
+					To:        v.Sender,
+					Type:      v.Type(),
+					Value:     tokenOutValues[1],
+				},
+				// token in (lp token)
+				{
+					Addresses: []string{v.Sender},
+					Origin:    v.Sender,
+					From:      v.Sender,
```
kaladinlight, line 84: We do have the `To` available here at `["transfer"]["recipient"]`

go/coinstacks/osmosis/tx.go
pastaghost: Going to write up a ticket for a fast follow on this one.

kaladinlight (approved):

PR #616, opened 2023-02-09
chore: adds page size validation to all UTXO and EVM coinstacks

0xean:
Adds basic validation to page size in UTXO and EVM coinstacks on transaction history endpoints

@kaladinlight - if you prefer to move the MAX_PAGE_SIZE var to somewhere else that could be modified without a code change, let me know.


kaladinlight (approved):
LGTM. I don't usually use doc comments outside of what is required for tsoa, but can leave them if you'd like

asamere (approved):

PR #617, opened 2023-02-09
fix: early return on bad page params for cosmos / osmosis

0xean:
Currently while we do validate the page params, if there is an issue, we continue to process the request. Some examples

https://dev-api.cosmos.shapeshift.com/api/v1/account/cosmos156gqf9837u7d4c4678yt3rl4ls9c5vuuxyhkw6/txs?pageSize=0
https://dev-api.cosmos.shapeshift.com/api/v1/account/cosmos156gqf9837u7d4c4678yt3rl4ls9c5vuuxyhkw6/txs?pageSize=101
https://dev-api.cosmos.shapeshift.com/api/v1/account/cosmos156gqf9837u7d4c4678yt3rl4ls9c5vuuxyhkw6/txs?pageSize=200

In all of these cases, the api does return an error, but still also proceeds to process (or attempt to process) the request with the given page size.

This PR fixes that issue with an early return to avoid attempting to process the request for cosmos and osmosis coin stacks.

It also appears this will fix thorchain as well

### DIFF
--- a/go/coinstacks/cosmos/api/api.go
+++ b/go/coinstacks/cosmos/api/api.go
@@ -135,7 +135,10 @@ func New(httpClient *cosmos.HTTPClient, grpcClient *cosmos.GRPCClient, wsClient 
 //	200: Validators
 //	500: InternalServerError
 func (a *API) GetValidators(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


asamere (approved):

kaladinlight (approved):
good catch, thanks!

PR #623, opened 2023-02-20
chore: upgrade geth to v1.11.1 and prysm to v3.2.1

kaladinlight:


gomesalexandre (approved):
Checked against upstream images, LGTM!

PR #626, opened 2023-02-21
Revert "chore: upgrade osmosis to v14.0.1"

kaladinlight:
Reverts shapeshift/unchained#624

- new version appears to introduce syncing problems, wait for next version to be published...


kaladinlight:
Seems to have picked up correctly on a fresh snapshot...

PR #630, opened 2023-02-27
fix: handle 0 decimal tokens

kaladinlight:
Various ERC20 tokens like [Atari](https://etherscan.io/token/0xdacD69347dE42baBfAEcD09dC88958378780FB62) actually have decimals set to 0 which were being filtered out. This change now returns any tokens that have at minimum, a balance and contract address.

Closes https://github.com/shapeshift/web/issues/3167


gomesalexandre (approved):
TIL 0/1 decimals is perfectly valid for ERC-20s, nice one ser

PR #638, opened 2023-03-08
[WIP] feat: Automatic EBS snapshots

### LINKED ISSUE #629: SPIKE: Automated Snapshots (EBS)
### Overview

To ensure there is a recent restore point in the case of corrupted db state or the need to scale/replicate, automated snapshots would be beneficial.

### References and additional details

https://aws.amazon.com/blogs/containers/using-amazon-ebs-snapshots-for-persistent-storage-with-your-amazon-eks-cluster-by-leveraging-add-ons/

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/examples/kubernetes/snapshot/README.md

### Acceptance Criteria

- Understand the necessary pre-requisites for supported automated snapshots in our cluster
- We use pulumi for our IAC. Determine what pulumi resources or custom resources/helm charts will be required to deploy
- Figure out the means for snapshot cleanup to ensure we don't burn $$$ with unnecessary snapshots
- Report findings and spec out concrete tasks for implementation

### Need By Date

_No response_

### Screenshots/Mockups

_No response_

### Estimated effort

_No response_

lmyslinski:
Related PR: https://github.com/shapeshift/cluster-launcher/pull/23
Closes #629

This PR introduced a set of changes which together add automatic EBS backup handling to each StatefulSet deployed via Unchained. Full set of changes includes:

- installation and configuration of https://github.com/kubernetes-csi/external-snapshotter via `cluster-launcher`
- a new CronJob optionally added to each `StatefulSet`
- a new component called `unchained backup runner`, which is trigged by the CronJob


lmyslinski:
[backup.log](https://github.com/shapeshift/unchained/files/10957339/backup.log)
Sample output from testing. I've tested a few different scenarios, this configuration is set to run a backup every minute and keep two historic backups. AWS console shows that we are stabily sitting at 8 snapshots as expected:
image

pulumi/backuprunner/.gitignore
```diff
@@ -0,0 +1,129 @@
+# Project specific
```
kaladinlight, line 1: think at this point it would just make sense to consolidate all `.gitignores` to a single top level file

pulumi/backuprunner/README.md
```diff
@@ -0,0 +1,30 @@
+
+# Unchained Backup Runner
+
+A small Node.js CLI tool for creationg k8s VolumeSnapshots for Unchained stateful sets.
```
kaladinlight, line 4:
```suggestion
A small Node.js CLI tool for creating k8s VolumeSnapshots for Unchained stateful sets.
```

pulumi/backuprunner/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+`ts-node-esm dist/index.js`
+
+## Build & Push a new version
+
+If you're build on M1 as I am, you need to do a cross-arch build:
+
+`docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v7 -t lukmyslinski/backuprunner:0.18 --push .`
+
+Once the version is pushed, update it in `pulumi/src/cluster/sts/backup-cron.ts`
```
kaladinlight, line 29: we can automate this in ci so it checks for any changes to source and pushes a new image to shapeshiftdao org

.gitignore
```diff
@@ -2,4 +2,5 @@
 .idea/
 node_modules/
 Pulumi.*.yaml
-!Pulumi.sample.yaml
\ No newline at end of file
+!Pulumi.sample.yaml
+dist/
```
kaladinlight, line 6: this is better than the copy, but I was actually thinking since we have a `.gitignore`, `node/.gitignore`, and `go/.gitignore`, just merging them all into `unchained/.gitignore`. i thought it might be nice to keep the rules separate, but I don't think there is actually any gain and just makes management more complicated

pulumi/backuprunner/src/backup.ts
```diff
@@ -0,0 +1,34 @@
+import k8s, { HttpError } from '@kubernetes/client-node'
+import { takeSnapshots } from './snapshot.js'
+import { scaleStatefulSet } from './scaling.js'
+import { cleanup } from './cleanup.js'
```
kaladinlight, line 4:
```suggestion
import { takeSnapshots } from './snapshot'
import { scaleStatefulSet } from './scaling'
import { cleanup } from './cleanup'
```

pulumi/backuprunner/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:18.12.1-alpine
```
kaladinlight, line 1: To align with all of our other dockerfiles, we can use the base image
```
ARG BASE_IMAGE

FROM $BASE_IMAGE
```

this will ensure we are using the same node/ts dependencies throughout the project

kaladinlight: This will require a bit of rework related to `node_modules` and workspace setup, so we can do a follow up

pulumi/backuprunner/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+COPY . .
+
+RUN yarn build
+
+WORKDIR dist
+
+ENTRYPOINT ["/usr/local/bin/node", "index.js", "$@"]
```
kaladinlight, line 9: we can also rely on ci for our build artifacts and just copy it in:

```
COPY . .

ENTRYPOINT ["node", "dist/index.js"]
```
no need for `$@` entrypoint args passed in via docker run

pulumi/backuprunner/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "type": "module",
+  "scripts": {
+    "start": "ts-node-esm dist/index.js",
+    "build": "npx tsc",
+    "dev": "ts-node-esm src/index.ts"
+  },
+  "dependencies": {
+    "@kubernetes/client-node": "^0.18.1",
+    "@types/figlet": "^1.5.5",
+    "@types/node": "^18.14.6",
```
kaladinlight, line 15: types can be moved to `devDependencies`

pulumi/backuprunner/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "commander": "^10.0.0",
+    "figlet": "^1.5.2",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.9.5"
```
kaladinlight, line 19: would like to keep these in sync with `node/package.json` for now. (we can probably upgrade `node/package.json`) to match so we are current

pulumi/backuprunner/package.json
kaladinlight, line 15: version sync here as well

pulumi/src/index.ts
```diff
@@ -27,15 +27,19 @@ export interface BaseConfig {
   rootDomainName?: string
 }
 
+export interface StsDefinition {
+  replicas: number
+  services: Array<ServiceConfig>,
+  backupSchedule?: string
+  backupCount?: number
```
kaladinlight, line 34:
```suggestion
  backup?: {
    schedule: string
    backupCount: number
  }
```

pulumi/src/index.ts
```diff
@@ -27,15 +27,19 @@ export interface BaseConfig {
   rootDomainName?: string
 }
 
+export interface StsDefinition {
```
kaladinlight, line 30:
```suggestion
export interface StatefulService {
```

pulumi/src/index.ts
```diff
@@ -27,15 +27,19 @@ export interface BaseConfig {
...
 export interface Config extends BaseConfig {
   stack: string
   network: string
   environment?: string
   api?: ApiConfig
-  statefulService?: {
-    replicas: number
-    services: Array<ServiceConfig>
-  }
+  statefulService?: StsDefinition
```
kaladinlight, line 42:
```suggestion
  statefulService?: StatefulService
```

pulumi/src/statefulService.ts
```diff
@@ -355,4 +356,8 @@ export async function deployStatefulService(
       { provider }
     )
   }
+
+  if(config.statefulService.backupCount && config.statefulService.backupSchedule){
```
kaladinlight, line 360:
```suggestion
  if (config.statefulService.backup) {
```

pulumi/src/statefulService.ts
```diff
@@ -355,4 +356,8 @@ export async function deployStatefulService(
...
+    deployStsBackupCron(asset, config.statefulService, namespace, provider);
```
kaladinlight, line 361: yeah we need to get formatting fixed up, i definitely botched things pulling out the shared pulumi code...
```suggestion
    deployStsBackupCron(asset, config.statefulService, namespace, provider)
```

pulumi/src/index.ts
kaladinlight: or `StatefulServiceConfig` if you like that better

pulumi/src/index.ts
kaladinlight: Will also want to add default/validation to `pulumi/src/config.ts` for new `backup` config

pulumi/src/statefulService.ts
kaladinlight: we should also only deploy backups only for sts in `unchained-dev`

pulumi/src/cluster/sts/backup-cron.ts
```diff
@@ -0,0 +1,102 @@
+import { ServiceConfig, StsDefinition } from "../..";
```
kaladinlight, line 1: might as well put this next to `statefulService.ts`, we can take a look together at re organizing things after we get this body of work in

pulumi/backuprunner/src/scaling.ts
```diff
@@ -0,0 +1,57 @@
...
+  k8sAppsClient: k8s.AppsV1Api,
+  name: string,
+  namespace: string,
+  count: number,
+  waitToFinish: boolean
+) => {
+  console.log(`Scaling StatefulSet ${namespace}.${name} to ${count} replicas`);
+
+  var res = await k8sAppsClient.readNamespacedStatefulSet(name, namespace);
+  let ss = res.body;
```
kaladinlight, line 17: prefer `const` over `let` or `var` throughout shapeshift stack

pulumi/backuprunner/src/scaling.ts
```diff
@@ -0,0 +1,57 @@
...
+  ss.spec!!.replicas = Number(count);
```
kaladinlight, line 18: if checks for type narrowing

pulumi/backuprunner/src/scaling.ts
```diff
@@ -0,0 +1,57 @@
...
+export const scaleStatefulSet = async (
+  k8sAppsClient: k8s.AppsV1Api,
+  name: string,
+  namespace: string,
+  count: number,
+  waitToFinish: boolean
+) => {
+  console.log(`Scaling StatefulSet ${namespace}.${name} to ${count} replicas`);
+
+  var res = await k8sAppsClient.readNamespacedStatefulSet(name, namespace);
```
kaladinlight, line 16: can also destructure
```suggestion
  const { body } = await k8sAppsClient.readNamespacedStatefulSet(name, namespace);
```

pulumi/backuprunner/README.md
```diff
@@ -0,0 +1,30 @@
...
+The tool is trigged via Cronjob in K8s. Based on the parameters, the following actions are performed:
+
+- The statefulset is scaled down to 0 
+- All PVC's claimed by the statefulset are snapshotted
+- The statefulset is scaled back up
+- All snapshots (VolumeClaims) over the configured number of backups are removed
+
+## Usage
+
+Run it locally or just see `index.ts`
```
kaladinlight, line 17: `yarn  start -h`

pulumi/backuprunner/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Run locally
+
+`ts-node-esm dist/index.js`
```
kaladinlight, line 21:
```suggestion
`yarn dev`
```

pulumi/backuprunner/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "backuprunner",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "type": "module",
+  "scripts": {
+    "start": "ts-node-esm dist/index.js",
+    "build": "npx tsc",
+    "dev": "ts-node-esm src/index.ts"
+  },
```
kaladinlight, line 11:
```suggestion
  "name": "@shapeshiftoss/unchained-backuprunner",
  "version": "1.0.0",
  "license": "MIT",
   "private": true,
  "scripts": {
    "clean": "rm -rf dist node_modules",
    "build": "tsc",
    "dev": "ts-node src/index.ts",
    "start": "node dist/index.js"
  },
```

kaladinlight: could also add prettier and a format command for sanity:
```json
    "format": "prettier -w --config ../../node/.prettierrc --ignore-path ../../node/.prettierignore ."
```

pulumi/backuprunner/src/scaling.ts
```diff
@@ -0,0 +1,63 @@
...
+) => {
+  console.log(`Scaling StatefulSet ${namespace}.${name} to ${count} replicas`);
+
+  const { body } = await k8sAppsClient.readNamespacedStatefulSet(name, namespace);
+
+  if(typeof(body.spec?.replicas) === "string"){
+    body.spec.replicas = count
+  }else{
+    console.warn(`Invalid replica count: ${body.spec?.replicas}`)
+  }
```
kaladinlight, line 22: I would go this route for narrowing:
```suggestion
  if (!body.spec) throw new Error(`No spec found for StatefulSet: ${namespace}.${name}`)

  body.spec.replicas = count
```
If we don't have a spec to modify, consider the scale/backup invalid. This lets ts know `spec` is no undefined, then we can just set `count` as desired.

`replicas` is `number | undefined` so `typeof string` should be invalid regardless

pulumi/backuprunner/src/scaling.ts
```diff
@@ -0,0 +1,63 @@
...
+  name: string,
+  namespace: string,
+  count: number
+) => {
+  const max_iterations = 100;
+
+  var done = false;
+  var iterations = 0;
+  
+  while (!done) {
```
kaladinlight, line 43: two thoughts here:
1) we could pass `terminationGracePeriod` in and just sleep for that amount of time, check state and proceed assuming it scaled down as expected within termination period (avoid looping/retry and just know it will take 2 mins to snapshot)
2) just use a for loop to simplify/reduce lets a bit:
```ts
  for (let i = 0; i < 100; i++) {
    // termination grace period is 120s so need to account for that
    console.log(`Waiting for ${name} to shut down...`)

    await delay(3000)
    const status = await k8sAppsClient.readNamespacedStatefulSetStatus(name, namespace)

    if (status.body.status?.availableReplicas === count) {
      console.log(`Scaling finished - ${namespace}.${name} availableReplicas is now ${count}`)
      return
    }
  }

  console.log(`Scaling StatefulSet ${namespace}.${name} has timed out`)
  process.exit(1)
```

pulumi/backuprunner/src/cleanup.ts
```diff
@@ -0,0 +1,56 @@
...
+  var pvcsToKeepCount = pvcCount*backupCount;
+
+  const snapshots = await getExistingVolumeSnapshots(k8sApi, sts, namespace);
+  console.log(`Backup count to keep - ${backupCount}. ${sts} consists of ${pvcCount} PVC's, so will keep ${pvcsToKeepCount} latest snapshots`)
+
+  if(snapshots.length <= pvcsToKeepCount){
+    console.log(`Not archiving old snapshots, too few results`)
+  }else {
+    await removeOldSnapshots(k8sApi, snapshots, pvcsToKeepCount)
+  }
```
kaladinlight, line 14:
```suggestion
  if (snapshots.length > pvcsToKeepCount) {
    await removeOldSnapshots(k8sApi, snapshots, pvcsToKeepCount)
  }
```

pulumi/backuprunner/src/cleanup.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const getExistingVolumeSnapshots = async (k8sApi: k8s.KubernetesObjectApi, sts: string, namespace: string): Promise<KubernetesObject[]> => {
+  const snapshots = await k8sApi.list("snapshot.storage.k8s.io/v1", "VolumeSnapshot", namespace, undefined, undefined, undefined, undefined, `statefulset=${sts}`);
+  const items = snapshots.body.items
+  console.log(`Found ${items.length} snapshots for ${namespace}.${sts}`)
+  return items.sort(sortByCreationTimestampDesc)
+}
+
+const sortByCreationTimestampDesc = ((a: KubernetesObject, b: KubernetesObject) => (new Date(b.metadata!!.creationTimestamp!!).getTime()  - new Date(a.metadata!!.creationTimestamp!!).getTime()));
```
kaladinlight, line 26: typically don't like to see `!!` but in this case I am ok since it is a very defined use case

pulumi/backuprunner/src/cleanup.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const removeOldSnapshots = async (k8sApi: k8s.KubernetesObjectApi, snapshots: KubernetesObject[], pvcsToKeepCount: number) => {
+  await Promise.all(getSnapshotsToRemove(snapshots, pvcsToKeepCount).map(async vc => await deleteSnapshot(k8sApi, vc)))
+}
+
+const deleteSnapshot = async (k8sApi: k8s.KubernetesObjectApi, vc: KubernetesObject) => {
```
kaladinlight, line 32: what is `vc` supposed to stand for? might be better to just name `snapshot`?

pulumi/backuprunner/src/cleanup.ts
```diff
@@ -0,0 +1,56 @@
+import k8s, { KubernetesObject } from '@kubernetes/client-node'
+
+export const cleanup = async (k8sApi: k8s.KubernetesObjectApi, sts: string, namespace: string, pvcList: string, backupCount: number) => {
+  var pvcCount = pvcList.split(',').length;
+  var pvcsToKeepCount = pvcCount*backupCount;
```
kaladinlight, line 5: feel like `snapshotsToKeepCount` would make a little more sense throughout

pulumi/src/backupCron.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+const getPvcNames = (asset: string, replicas: number, services: ServiceConfig[]) => {
+  return (Array.from(Array(replicas).keys()).flatMap(n => {
+      return services.map(svc => `data-${svc.name}-${asset}-sts-${n}`)
+    }).filter((pvc) => pvc) as string[]).join(',')
+}
+
+
+const createRbac = (asset: string, namespace: string, provider: k8s.Provider) => {
```
kaladinlight, line 62: feels like we should be able to get the service account created by the snapshot controller for appropriate perms. could either pass back from cluster-launcher and assign as a stack reference output, or make that configurable  through cluster launcher and just pass that config down to stateful service -> backupCron

pulumi/src/backupCron.ts
```diff
@@ -0,0 +1,110 @@
+import { ServiceConfig, StatefulService } from ".";
+import * as k8s from '@pulumi/kubernetes'
+
+export const deployStsBackupCron = (asset: string, sts: StatefulService, namespace: string, provider: k8s.Provider) => {
```
kaladinlight, line 4: breaking things out into separate functions has it's uses, but depending on the types accepted by those functions, you can lose useful type narrowing along the way. I would personally just have a single func (assuming we can get serviceAccountName) as follows:

```ts
export const deployStsBackupCron = (asset: string, sts: StatefulService, namespace: string, provider: k8s.Provider) => {    
  if (!sts.backup) return

  const pvcs = sts.services.map((svc, i) => `data-${svc.name}-${asset}-sts-${i}`).join(',')

  const backupContainer = {
    name: `${asset}-backup-runner`,
    image: 'lukmyslinski/backuprunner:0.24',
    args: ['-n', namespace, '-s', `${asset}-sts`, '-p', pvcs, '-r', `${sts.replicas}`, "-c", `${sts.backup.backupCount}`],
  }

  new k8s.batch.v1.CronJob(`${asset}-backup-job`, {
    metadata: {
      name: `${asset}-backup-job`,
      namespace: namespace,
      annotations: { 'pulumi.com/skipAwait': 'true' },
    },
    spec: {
      successfulJobsHistoryLimit: 1,
      failedJobsHistoryLimit: 1,
      concurrencyPolicy: "Forbid",
      schedule: sts.backup.schedule,
      jobTemplate: {
        spec: {
          template: {
            spec: {
              serviceAccountName: 'passThroughArg',
              containers: [backupContainer],
              restartPolicy: "Never"
            },
          },
        },
      },
    },
  }, { provider })
}
```

kaladinlight: no need for `!!` or `?` as we have already narrowed the type and retain that scope.

we will need to build out `sts.backup` a bit more. we can either require `backupCount` and `schedule` in which we will want to validate in `pulumi/config.ts`, or we can change those to optional types and provide defaults internally:

```ts
  if (!sts.backup) return

  const { backupCount = 1, schedule = '0 4 * * 0' } = sts.backup
  
  OR

  const backupCount = sts.backup.backupCount ?? 1
  const schedule = sts.backup.schedule ?? '0 4 * * 0' 
```

kaladinlight:
one more comment related to how we are handling scaling and the pvc list. we should only have to scale sts down by 1 and snapshot the highest index.

targetReplicas = currentReplicas -1
pvcList = services.map(svc => `${name}-${currentReplicas-1}`).join(',')

this way we don't have to pass in replicas anymore and only scale down by one, and take the snapshot state of a single replica

lmyslinski:
> one more comment related to how we are handling scaling and the pvc list. we should only have to scale sts down by 1 and snapshot the highest index.
> 
> targetReplicas = currentReplicas -1 pvcList = services.map(svc => `${name}-${currentReplicas-1}`).join(',')
> 
> this way we don't have to pass in replicas anymore and only scale down by one, and take the snapshot state of a single replica

That's a great hint

pulumi/src/backupCron.ts
lmyslinski: I see, definitely makes sense to do so. I'm more accustomed to stricter SRP coming mostly from the JVM land but having fewer types to verify is definitely a huge gain here

pulumi/src/backupCron.ts
lmyslinski: This is the only point I'm not that sure of. I'd say it's better if we keep a single domain object defined in a single place/repo, otherwise doing any tweaks will require more back and forth. Had those been more tightly coupled I'd probably agree, but the cronjob is currently totally encapsulated here and I see more hard than good in splitting it up

pulumi/backuprunner/src/snapshot.ts
```diff
@@ -0,0 +1,40 @@
+import k8s, { KubernetesObject } from '@kubernetes/client-node';
+
+interface VolumeSnapshot extends KubernetesObject {
+  spec: {
+    volumeSnapshotClassName: string,
+    source: {
+      persistentVolumeClaimName: string
+    }
+  }
+}
```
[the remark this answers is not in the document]
lmyslinski: I really needed this comment, thanks

pulumi/backuprunner/src/cleanup.ts
lmyslinski: should've been a `vs`, `volumeSnapshot`. Will patch

pulumi/backuprunner/src/cleanup.ts
lmyslinski: Agreed, couldn't find a better way but with the change to the interface from down below it might be avoidable

pulumi/backuprunner/src/scaling.ts
lmyslinski: Even though the code is dumb and ugly, I'm a fan of simplicity. There's no way that code breaks due to some weird k8s race condition 2 releases down the road. It would definitely be nicer to pass the value down, but it's slightly more complex and prone to misconfiguration. In the 2nd case we either wait far too long or run into the same issue - there's a need to tweak the grace period. 

Ultimately I'd prefer keeping it as is or alternatively pulling in some util library to handle this for us, as there's very little room for error here and it minimizes downtime. Maybe just some simply exponential backoff like https://www.npmjs.com/package/exponential-backoff?

pulumi/backuprunner/src/scaling.ts
lmyslinski: Missed the wrong type check, thx

pulumi/backuprunner/src/cleanup.ts
kaladinlight: Yeah I tested it out locally and that should give us the explicit typing needed to avoid this. Thought about it after this comment haha

pulumi/backuprunner/src/cleanup.ts
kaladinlight: I would say generally speaking, our typescript variable naming is more verbose/explicit versus the golang shorthand. Even with the golang shorthand, it is just iterators or narrowly scoped variables. With the shorthand param and being passed through several functions/callbacks, it loses it's meaning a bit regardless imo

pulumi/backuprunner/src/scaling.ts
kaladinlight: I can get behind the bullet proof retry loop over grace period config. What makes you think that about the second case? Should be the exact same logic as your while loop, just as a for loop to reduce some extra variable overhead.

pulumi/src/backupCron.ts
kaladinlight: And just to clarify, does the config validation make sense regarding the backup config? Let me know if not and I can walk through a bit more, might be easier to do a quick screen share than try to explain via text.

pulumi/src/backupCron.ts
kaladinlight: I can see your point to an extent about the encapsulation, but it also feels like we are missing out on leveraging a decent sized chunk of infra established by the helm chart. By exposing the service account through the stack output, it would guarantee that any changes made to the cluster launcher config would be associated via stack reference creating and "automated" link in a sense. Want to have a quick chat in discord about this just to prove both side out a bit further?

pulumi/backuprunner/src/scaling.ts
lmyslinski: didn't notice the return, I see it now, all clear

pulumi/backuprunner/src/cleanup.ts
lmyslinski: ...let's definitely avoid the go naming patterns :)

pulumi/src/backupCron.ts
lmyslinski: Tbh I gave it another thought and yeah, let's just pull it up. It's a service account, shouldn't be touched very often

pulumi/src/backupCron.ts
lmyslinski: You mean the scoped type narrowing via `if` guards? All clear

lmyslinski: It does enforce a certain code structure but it's liveable. Not as good as i.e. a monadic approach but hey, it's simple enough

pulumi/src/backupCron.ts
kaladinlight: as discussed in chat, leaving as is due to namespaced service accounts not allowing us to leverage `unchained-infra` role bindings/service accounts

pulumi/src/backupCron.ts
kaladinlight: more from the pulumi config side. see: https://github.com/shapeshift/unchained/blob/d13399f8a1d3f782f891bbde423cd5a3410c3d88/pulumi/src/config.ts

`new Pulumi.Config` does not enforce types unfortunately, so we have extra validation in there to basically say "if this optional object is defined, ensure all required fields actually exists" so we don't end up with an `undefined` on something we think should be required based on the type def

pulumi/src/cluster/index.ts
```diff
@@ -100,6 +100,23 @@ export = async (): Promise<Outputs> => {
         cacheFroms: [`${blockbookImage}:${blockbookTag}`, `${blockbookImage}:latest`],
       })
     }
+
+    const volumeReaperImage = `${config.dockerhub.username}/volumereaper`
```
kaladinlight, line 104:
```suggestion
    const volumeReaperImage = `${config.dockerhub.username}/${name}-volume-reaper`
```

pulumi/src/cluster/index.ts
```diff
@@ -100,6 +100,23 @@ export = async (): Promise<Outputs> => {
...
+      image: volumeReaperImage,
+      context: '../../volumeReaper',
+      auth: {
+        password: config.dockerhub.password,
+        username: config.dockerhub.username,
+        server: config.dockerhub.server,
+      },
+      buildArgs: buildArgs,
+      env: { DOCKER_BUILDKIT: '1' },
+      tags: ['test-hash'],
```
kaladinlight, line 117: ideally we folder hash the contents of `volumeReaper` and only build and push if there are changes, using hash as the tag. you can see a reference here: https://github.com/shapeshift/unchained/blob/72f0eccfdf5ecd579567619b92426015adcf0335/pulumi/src/hasher.ts

pulumi/src/cluster/index.ts
```diff
@@ -100,6 +100,23 @@ export = async (): Promise<Outputs> => {
...
+      cacheFroms: [`${volumeReaperImage}:${baseTag}`, `${volumeReaperImage}:latest`],
```
kaladinlight, line 118: can also add `${baseImage}:latest` to cache from

pulumi/volumeReaper/Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG BASE_IMAGE
+
+FROM $BASE_IMAGE
+
+COPY . .
+
+WORKDIR dist
+
+ENTRYPOINT ["/usr/local/bin/node", "index.js", "$@"]
```
kaladinlight, line 9:
```suggestion

ENTRYPOINT ["node", "dist/index.js"]
```

pulumi/volumeReaper/README.md
```diff
@@ -0,0 +1,30 @@
...
+`yarn dev`
+
+## Build & Push a new version
+
+If you're build on M1 as I am, you need to do a cross-arch build:
+
+`docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v7 -t lukmyslinski/backuprunner:0.18 --push .`
+
+Once the version is pushed, update it in `pulumi/src/cluster/sts/backup-cron.ts`
+
```
kaladinlight, line 30: not sure if super relevant with auto build in ci
```suggestion
```

pulumi/volumeReaper/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "format": "prettier -w --config ../../node/.prettierrc --ignore-path ../../node/.prettierignore .",
+    "start": "node dist/index.js"
+  },
+  "dependencies": {
+    "@kubernetes/client-node": "^0.18.1",
+    "commander": "^10.0.0",
+    "figlet": "^1.5.2",
+    "prettier": "^2.8.4",
+    "ts-node": "^10.8.1",
+    "typescript": "^4.7.4"
```
kaladinlight, line 19: these can be moved to `devDependencies` as well. Rule of thumb is all types or any packages only used during local dev (not runtime)

pulumi/volumeReaper/src/backup.ts
```diff
@@ -0,0 +1,36 @@
...
+  const k8sObjectClient = k8s.KubernetesObjectApi.makeApiClient(kc)
+  const k8sAppsClient = kc.makeApiClient(k8s.AppsV1Api)
+  const replicas = await getCurrentReplicas(k8sAppsClient, opts.statefulset, opts.namespace)
+
+  try {
+    await scaleStatefulSet(k8sAppsClient, opts.statefulset, opts.namespace, replicas - 1, true)
+    await takeSnapshots(k8sObjectClient, opts.statefulset, opts.pvcList)
+    await scaleStatefulSet(k8sAppsClient, opts.statefulset, opts.namespace, replicas, false)
+    await cleanup(k8sObjectClient, opts.statefulset, opts.namespace, opts.pvcList, opts.backupCount)
+    console.log('Backup runner completed')
```
kaladinlight, line 28:
```suggestion
    console.log('Backup completed')
```

pulumi/volumeReaper/src/index.ts
```diff
@@ -0,0 +1,29 @@
+import figlet from 'figlet'
+import { Command } from 'commander'
+import { runBackup } from './backup'
+
+console.log(figlet.textSync('Unchained Backup Runner'))
```
kaladinlight, line 5:
```suggestion
console.log(figlet.textSync('Unchained Volume Reaper'))
```

pulumi/volumeReaper/src/snapshot.ts
```diff
@@ -0,0 +1,45 @@
+import k8s, { KubernetesObject } from '@kubernetes/client-node'
```
kaladinlight, line 1: `import * as` update

pulumi/volumeReaper/src/scaling.ts
```diff
@@ -0,0 +1,50 @@
+import k8s from '@kubernetes/client-node'
```
kaladinlight, line 1:
```suggestion
import * as k8s from '@kubernetes/client-node'
```

pulumi/src/cluster/docker.ts
```diff
@@ -0,0 +1,64 @@
...
+      context: '../../volumeReaper',
+      auth: {
+        password: dockerhub.password,
+        username: dockerhub.username,
+        server: dockerhub.server,
+      },
+      buildArgs: { BUILDKIT_INLINE_CACHE: '1', BASE_IMAGE: baseImage },
+      env: { DOCKER_BUILDKIT: '1' },
+      tags: [volumeReaperHash],
+      cacheFroms: [`${baseImage}:latest`, `${volumeReaperImage}:latest`, `${volumeReaperImage}:${volumeReaperHash}`,],
```
kaladinlight, line 61: order from most specific to least specific images, first one hit will be used, so in this case the base image would be used and produce the least complete cache state

pulumi/src/cluster/docker.ts
```diff
@@ -0,0 +1,64 @@
+import { hashElement } from "folder-hash"
+import { Dockerhub } from ".."
+import { hasTag, buildAndPushImage } from "../docker"
+import { getBaseHash, getVolumeReaperHJash as getVolumeReaperHash } from "../hasher"
+
+
+export const buildAndPushDockerImages = async (dockerhub: Dockerhub) => {
```
kaladinlight, line 7: thanks for taking the time to separate some of this stuff out to make it more readable/maintainable

pulumi/src/hasher.ts
```diff
@@ -37,3 +38,14 @@ export const getBaseHash = async (): Promise<string> => {
 
   return hash.digest('hex')
 }
+
+export const getVolumeReaperHJash = async (): Promise<string> => {
```
kaladinlight, line 42:
```suggestion
export const getVolumeReaperHash = async (): Promise<string> => {
```

pulumi/src/hasher.ts
```diff
@@ -37,3 +38,14 @@ export const getBaseHash = async (): Promise<string> => {
 
   return hash.digest('hex')
 }
+
+export const getVolumeReaperHash = async (): Promise<string> => {
+  const hash = createHash('sha1')
+  // hash volume reaper files
+  const { hash: volumeReaperHash } = await hashElement(volumeReaperDir, {
+    folders: { include: ['**'], exclude: ['.*', 'dist', 'node_modules'] },
+    files: { include: ['package.json'] },
```
kaladinlight, line 47:
```suggestion
    files: { include: ['*.ts', '*.json', 'Dockerfile'] },
```

pulumi/volumeReaper/src/snapshot.ts
```diff
@@ -0,0 +1,47 @@
+import * as k8s from '@kubernetes/client-node'
+import { KubernetesObject } from '@kubernetes/client-node'
```
kaladinlight, line 2: we can always use `k8s.KubernetesObject` to avoid multiple imports of the same package. would chose one or the other, and stick with that for this package throughout.

pulumi/volumeReaper/src/backup.ts
```diff
@@ -0,0 +1,44 @@
...
+  kc.loadFromDefault()
+  const k8sObjectClient = k8s.KubernetesObjectApi.makeApiClient(kc)
+  const k8sAppsClient = kc.makeApiClient(k8s.AppsV1Api)
+
+  const statefulset = `${opts.asset}-sts`
+
+  const replicas = await getCurrentReplicas(k8sAppsClient, statefulset, opts.namespace)
+  assert(replicas >= 1, 'Replicas needs to be larger than 0 to run backup')
+
+  const pvcList = opts.stsServices.split(',').map((svc) => `data-${svc}-${opts.asset}-sts-${replicas}`)
```
kaladinlight, line 28:
```suggestion
  const pvcList = opts.stsServices.split(',').map((svc) => `data-${svc}-${opts.asset}-sts-${replicas - 1}`)
```

pulumi/volumeReaper/src/backup.ts
```diff
@@ -0,0 +1,44 @@
+import * as k8s from '@kubernetes/client-node'
+
+import { takeSnapshots } from './snapshot'
+import { getCurrentReplicas, scaleStatefulSet } from './scaling'
+import { cleanup } from './cleanup'
+import { HttpError } from '@kubernetes/client-node'
```
kaladinlight, line 6: same import comment

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
+import k8s, { KubernetesObject } from '@kubernetes/client-node'
```
kaladinlight, line 1: same import comment

pulumi/volumeReaper/src/scaling.ts
```diff
@@ -0,0 +1,50 @@
+import * as k8s from '@kubernetes/client-node'
+
+export const getCurrentReplicas = async (
+  k8sAppsClient: k8s.AppsV1Api,
+  name: string,
+  namespace: string
+): Promise<number> => {
+  const { body } = await k8sAppsClient.readNamespacedStatefulSetStatus(name, namespace)
+  if (!body.spec?.replicas) throw new Error(`No spec found for StatefulSet: ${namespace}.${name}`)
+  return body.spec?.replicas
```
kaladinlight, line 10:
```suggestion
  return body.spec.replicas
```

pulumi/volumeReaper/src/snapshot.ts
```diff
@@ -0,0 +1,47 @@
...
+  console.log(`Taking snapshot of pvc ${pvcName} - ${snapshotName}`)
+  const snapshotYaml: VolumeSnapshot = {
+    apiVersion: 'snapshot.storage.k8s.io/v1',
+    kind: 'VolumeSnapshot',
+    metadata: {
+      name: snapshotName,
+      labels: {
+        statefulset: sts,
+      },
+      creationTimestamp: new Date(), // will be overwritten by k8s
```
kaladinlight, line 36: good comment, was wondering if it would take a defined timestamp or overwrite

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+    undefined,
+    `statefulset=${sts}`
+  )
+  const items = snapshots.body.items
+  console.log(`Found ${items.length} snapshots for ${namespace}.${sts}`)
+  return items.sort(sortByCreationTimestampDesc)
+}
+
+const sortByCreationTimestampDesc = (a: VolumeSnapshot, b: VolumeSnapshot) =>
+  new Date(b.metadata?.creationTimestamp).getTime() - new Date(a.metadata?.creationTimestamp).getTime()
```
kaladinlight, line 39: same here, now that we are using `VolumeSnapshot`
```suggestion
  new Date(b.metadata.creationTimestamp).getTime() - new Date(a.metadata.creationTimestamp).getTime()
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const removeOldSnapshots = async (
+  k8sApi: k8s.KubernetesObjectApi,
+  snapshots: KubernetesObject[],
+  snapshotsToKeepCount: number
+) => {
+  await Promise.all(
+    getSnapshotsToRemove(snapshots, snapshotsToKeepCount).map(async (vc) => await deleteSnapshot(k8sApi, vc))
```
kaladinlight, line 47:
```suggestion
    getSnapshotsToRemove(snapshots, snapshotsToKeepCount).map(async (snapshot) => await deleteSnapshot(k8sApi, snapshot))
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+  console.log(`Found ${items.length} snapshots for ${namespace}.${sts}`)
+  return items.sort(sortByCreationTimestampDesc)
+}
+
+const sortByCreationTimestampDesc = (a: VolumeSnapshot, b: VolumeSnapshot) =>
+  new Date(b.metadata?.creationTimestamp).getTime() - new Date(a.metadata?.creationTimestamp).getTime()
+
+const removeOldSnapshots = async (
+  k8sApi: k8s.KubernetesObjectApi,
+  snapshots: KubernetesObject[],
```
kaladinlight, line 43: fully plumb through `VolumeSnapshot`
```suggestion
  snapshots: VolumeSnapshot[],
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+  snapshotsToKeepCount: number
+) => {
+  await Promise.all(
+    getSnapshotsToRemove(snapshots, snapshotsToKeepCount).map(async (vc) => await deleteSnapshot(k8sApi, vc))
+  )
+}
+
+const deleteSnapshot = async (k8sApi: k8s.KubernetesObjectApi, snapshot: KubernetesObject) => {
```
kaladinlight, line 51:
```suggestion
const deleteSnapshot = async (k8sApi: k8s.KubernetesObjectApi, snapshot: VolumeSnapshot) => {
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+  await k8sApi.delete({
+    apiVersion: snapshot.apiVersion,
+    kind: snapshot.kind,
+    metadata: {
+      name: snapshot.metadata?.name,
+      labels: snapshot.metadata?.labels,
+    },
+  })
+  console.log(`Deleted ${snapshot.metadata?.name}`)
```
kaladinlight, line 60:
```suggestion
      name: snapshot.metadata.name,
      labels: snapshot.metadata.labels,
    },
  })
  console.log(`Deleted ${snapshot.metadata.name}`)
```

kaladinlight: Could really just pass in `snapshot` thinking about it.
```ts
await k8sApi.delete(snapshot)
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const getSnapshotsToRemove = (snapshots: KubernetesObject[], pvcsToKeepCount: number) => {
```
kaladinlight, line 63:
```suggestion
const getSnapshotsToRemove = (snapshots: VolumeSnapshot[], pvcsToKeepCount: number) => {
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+  const [toKeep, toRemove] = [snapshots.slice(0, pvcsToKeepCount), snapshots.slice(pvcsToKeepCount)]
+
+  console.log(`Volume snapshots to keep: ${toKeep.length}`)
+  toKeep.forEach((x) => console.log(x.metadata?.name))
+
+  console.log(`Volume snapshots to remove: ${toRemove.length}`)
+  toRemove.forEach((x) => console.log(x.metadata?.name))
```
kaladinlight, line 70:
```suggestion
  toKeep.forEach((x) => console.log(x.metadata.name))

  console.log(`Volume snapshots to remove: ${toRemove.length}`)
  toRemove.forEach((x) => console.log(x.metadata.name))
```

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  // Start removal from the oldest VC's
```
kaladinlight, line 72:
```suggestion
  // Start removal from the oldest snapshot first
```

pulumi/src/cluster/docker.ts
```diff
@@ -0,0 +1,64 @@
+import { hashElement } from "folder-hash"
+import { Dockerhub } from ".."
+import { hasTag, buildAndPushImage } from "../docker"
+import { getBaseHash, getVolumeReaperHJash as getVolumeReaperHash } from "../hasher"
```
kaladinlight, line 4: don't forget this too
```suggestion
import { getBaseHash, getVolumeReaperHash } from "../hasher"
```

pulumi/src/cluster/docker.ts
```diff
@@ -0,0 +1,64 @@
...
+
+
```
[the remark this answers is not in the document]
kaladinlight: this will actually resolve if you want to run:
```bash
prettier -w --config ../../node/.prettierrc --ignore-path ../../node/.prettierignore ../src/
```
from `unchained/pulumi/volumeReaper`

pulumi/volumeReaper/src/cleanup.ts
lmyslinski: Good catch, I've added the type all over the file

pulumi/volumeReaper/src/cleanup.ts
```diff
@@ -0,0 +1,74 @@
+import k8s from '@kubernetes/client-node'
```
kaladinlight, line 1: don't want to run into same default import error case
```suggestion
import * as k8s from '@kubernetes/client-node'
```

kaladinlight: Interesting... well at least they all follow the same pattern now I guess haha

PR #665, opened 2023-03-24
fix: volume reaper 409 problem on k8s sts scaling

lmyslinski:


pulumi/volumeReaper/src/volumeReaper.ts
```diff
@@ -92,10 +92,11 @@ export class VolumeReaper {
     return sts
   }
 
-  private async scaleStatefulSet(_sts: StatefulSet, count: number, skipAwait: boolean): Promise<void> {
+  private async scaleStatefulSet(count: number, skipAwait: boolean): Promise<void> {
+
```
kaladinlight, line 96: nitpick
```suggestion
```

kaladinlight (approved):

PR #670, opened 2023-03-28
feat: naming unification

lmyslinski:
The problem:

It is currently impossible to reliably test non-default coinstack configuration, as the some names are propagated downstream in multiple ways and require certain values to be set to defaults in order to work. 

Examples include:

`const [coinstack] = asset.split('-')`
`const { hash: apiHash } = await hashElement(`../../${coinstack}/api`, `
`const name = `${args.asset}-${args.config.name}`

The combination of these makes it impossible to change the name of deployed resources. Also the terms `name`, `asset`, `service` are used inconsitently, i.e. a parameter passed down as `name` is received as `app` downstream, further complicating debugging of naming issues.

I'd like to introduce a shared top-down naming scheme that will make it obvious which resources will be referenced, where, and under what name. 

Sample updated config:

```
config:
  unchained:coinstack:
    assetName: litecointest
    environment: dev
    stack: unchained/common/dependencies-us-east-2
    network: mainnet
```

The hierarchy would be as follows (top to bottom):

`appName -> assetName -> serviceName`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -841,41 +841,41 @@ jobs:
           working_directory: node/<< parameters.pulumi-dir >>
           command: |
             pulumi stack select $<< parameters.organization >>_PULUMI_ORG/<< parameters.pulumi-stack >>
-            pulumi config set --path unchained.coinstack.assetName << parameters.assetName >>
-            pulumi config set --path unchained.coinstack.stack $<< parameters.organization >>_PULUMI_ORG/common/dependencies-us-east-2
-            [ -n "<< parameters.environment >>" ] && pulumi config set --path unchained.coinstack.environment << parameters.environment >>
-            pulumi config set --path unchained.coinstack.network << parameters.network >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.autoscaling.enabled << parameters.api-autoscaling >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.autoscaling.cpuThreshold << parameters.api-cpu-threshold >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.autoscaling.maxReplicas << parameters.api-max-replicas >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.cpuLimit << parameters.api-cpu-limit >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.cpuRequest << parameters.api-cpu-request >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.memoryLimit << parameters.api-memory-limit >>
-            [ "<< parameters.api >>" = true ] && pulumi config set --path unchained.coinstack.api.replicas << parameters.api-replicas >>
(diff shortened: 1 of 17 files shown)


.circleci/config.yml
```diff
@@ -841,40 +841,41 @@ jobs:
...
-            [ << parameters.stateful-service-replicas >> -gt 0 ] && [ -n "<< parameters.service-cpu-request-3 >>" ] && pulumi config set --path unchained:<< parameters.coinstack >>.statefulService.services[2].cpuRequest << parameters.service-cpu-request-3 >>
-            [ << parameters.stateful-service-replicas >> -gt 0 ] && [ -n "<< parameters.service-memory-limit-3 >>" ] && pulumi config set --path unchained:<< parameters.coinstack >>.statefulService.services[2].memoryLimit << parameters.service-memory-limit-3 >>
-            [ << parameters.stateful-service-replicas >> -gt 0 ] && [ -n "<< parameters.service-memory-request-3 >>" ] && pulumi config set --path unchained:<< parameters.coinstack >>.statefulService.services[2].memoryRequest << parameters.service-memory-request-3 >>
-            [ << parameters.stateful-service-replicas >> -gt 0 ] && [ -n "<< parameters.service-storage-size-3 >>" ] && pulumi config set --path unchained:<< parameters.coinstack >>.statefulService.services[2].storageSize << parameters.service-storage-size-3 >>
+            pulumi config set --path unchained.coinstack.assetName << parameters.assetName >>
```
kaladinlight, line 844: pretty sure all of the `<< parameters.coinstack >>` -> `coinstack` changes need to use the `:` for the correct syntax
```suggestion
            pulumi config set --path unchained:coinstack.assetName << parameters.assetName >>
```

pulumi/src/api.ts
```diff
@@ -48,15 +50,14 @@ export async function deployApi(args: DeployApiArgs): Promise<k8s.apps.v1.Deploy
   if (config.api === undefined) return
 
   const tier = 'api'
-  const labels = { app, asset, tier }
-  const name = `${asset}-${tier}`
-  const [coinstack] = asset.split('-')
+  const labels = { coinstack, assetName, tier }
```
kaladinlight, line 53: we lost `app` label. also `app` -> `appName` and `asset` -> `assetName` create a diff across all api resources and no longer match labels in `statefulService`. I don't find the `Name` suffix super useful, but I will let you make the call there and make sure all labels reflect the same naming convention with this update

kaladinlight:
few comments to address and will need to update all of the coinstacks to match. good improvement!

pulumi/src/api.ts
lmyslinski: On the k8s level I agree it's pointless, but on TS level I do find it useful to differentiate objects from strings. I've aligned it so that we only modify the TS vars

.circleci/config.yml
lmyslinski: 👍

pulumi/src/api.ts
kaladinlight: I typically just rely on the associated type, but I'm ok with leaving for more explicit clarity

pulumi/src/api.ts
```diff
@@ -17,11 +17,12 @@ export interface ApiConfig {
 }
 
 export interface DeployApiArgs {
-  app: string
-  asset: string
+  appName: string
+  coinstack: string
+  assetName: string
```
kaladinlight, line 22: want to make the same naming/label changes for `statefulService` as well?

lmyslinski: sure

kaladinlight (approved):
:fire:

PR #672, opened 2023-03-30
fix: Always build x64 docker images

lmyslinski:
When images are built locally on M1 macs, they're built and pushed with aarch64 architecture, which is incompatible with our nodes. This additional argument has no effect on x64 systems, but builds correct images when on a different architecture

### DIFF
--- a/pulumi/src/docker.ts
+++ b/pulumi/src/docker.ts
@@ -59,7 +59,7 @@ export const buildAndPushImage = async (args: BuildAndPushImageArgs): Promise<vo
 
   dockerBuildArgs.push(args.context)
 
-  execSync(`${envs} docker build ${dockerBuildArgs.join(' ')}`, { stdio: 'inherit' })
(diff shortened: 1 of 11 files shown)


pulumi/src/docker.ts
```diff
@@ -59,7 +59,7 @@ export const buildAndPushImage = async (args: BuildAndPushImageArgs): Promise<vo
 
   dockerBuildArgs.push(args.context)
 
-  execSync(`${envs} docker build ${dockerBuildArgs.join(' ')}`, { stdio: 'inherit' })
+  execSync(`${envs} docker build --platform linux/amd64 ${dockerBuildArgs.join(' ')}`, { stdio: 'inherit' })
```
kaladinlight, line 62: want to hit all of the other `docker run` or `docker build` commands throughout with the same treatment?

lmyslinski: Ah yeah forgot there was more

kaladinlight (approved):

PR #675, opened 2023-04-03
chore: lift node project config to root directory

kaladinlight:
there was complication around hoisted node modules, lint rules, build command throughout the project for shared node resources. this updates to bring node setup to the root directory so the whole project is nested under the same workspace, shares the same node_modules/yarn.lock, and node install/build commands. this also helps IDEs since they expect to find linting/formatting config at project root.

a follow up would be to lift golang project config to root directory as well so everything is controlled from the root and only coinstacks/language specific shared logic are nested under `unchained/node` and `unchained/go`.


pulumi/src/snapper.ts
```diff
@@ -0,0 +1,132 @@
...
+export interface SnapperArgs {
+  assetName: string
+  kubeconfig?: string
+  namespace: string
+}
+
+export class Snapper {
+  readonly assetName: string
+  readonly stsName: string
+  readonly namespace: string
```
lmyslinski, line 31: Good to have this bit unified

lmyslinski:
I pretty much agree with everything said here, I'd add two additional great next steps:

- Separating the Pulumi source tree from the coinstack implementations
- Extracing and sharing the hashing functions etc. so that we're using as little relative imports and paths as possible

lmyslinski (approved):

kaladinlight:
> I pretty much agree with everything said here, I'd add two additional great next steps:
> 
> * Separating the Pulumi source tree from the coinstack implementations
> * Extracing and sharing the hashing functions etc. so that we're using as little relative imports and paths as possible

Yep, I like the sounds of those

PR #677, opened 2023-04-04
refactor: coinstack unify

lmyslinski:
The mother of all refactors is here.

This PR unifies the configuration logic for all of our coinstacks - as you can see, a lot of our logic is currently duplicated across all coinstacks, with minor tweaks between them. I believe I've managed to extract all the common parts while not loosing any of the configuration options. 

Sadly there isn't a good way to do this partially, as everything is tightly coupled here (except maybe for the hash stuff)

The PR consists of the following changes:
- The logic for calculating the hash of each of the coinstack files has been centralized under the Pulumi source directory. 
- The process for configuring the coinstack has been separated into more distinct puzzles and workflows, which should allow for less intrusive changes in the future. The new process can be summarized as following:

- The starting point for each coinstack is the `index.ts` in the `pulumi` dir of each coinstack
- The `index.ts` contains configuration options for the coinstack which are not / cannot be expressed in the Pulumi config
- The `index.ts` also defines each of the `coinservices`, which are containers that make up the `statefulset` of each coinstack
- Once inside `deployCoinstack`, the input from the Pulumi config as well as the tweaks from `index.ts` are merged into one large typed input object called `CoinServiceInput`. 
- Validations are performed on the input object
- The unchained API for each coinstack is deployed
- The coinstack stateful set is aggregated from all the `CoinServiceInput`'s. This part contains one ugly bit called `enrichWithPulumiConfig`, which performs some one off-tweaks. Because these configs require access to validated config, they don't fit the flow very well and need to be performed here. 
- Finally, once the STS has been assembled, the sts is deployed, completing the flow

Testing considerations:
- For testing purposes, CircleCI has been tweaked to preview the changes this PR will introduce. These changes will be reverted once approved
- Individual coinstacks should be deployed under different names and manually inspected to catch any potential errors.


pulumi/src/api.ts
```diff
@@ -18,33 +19,51 @@ export interface ApiConfig {
...
+  const nodeBasePath = `../../node/${coinstack}`
+  const goBasePath = `../../go/${coinstack}`
+
+  switch(coinstackType){
+    case CoinstackType.GO:  
+      return { context: `${goBasePath}/go`, dockerFile: `${goBasePath}/build/Dockerfile` }
+    case CoinstackType.NODE:
+      return { context: `${nodeBasePath}/api` }
+  }
+}
```
lmyslinski, line 44: The big problem I still see is having these path related code bits all over the place. Still trying to figure out the best way to solve that

lmyslinski:
So far I think I've managed to abstract most of the complexity and am quite happy with the result, but there is one edge case that totally ruins this concept: injecting env vars that depend on pulumi output. 

As far as I've seen some can be skipped (ex. the `NETWORK` vars for various chains, I didn't see such a param having an effect on the underlying docker images per their docs) but for others it's definitely required like optimism or thorchain:

`env: { L1_RPC_ENDPOINT: `http://ethereum-svc.${namespace}.svc.cluster.local:8332` },`
or 
`env: { CHAIN_ID: `${coinstack}-${config.network}-v1`, NET: config.network },`

I don't have a good idea yet on how to address this, open to suggestions

lmyslinski:
Also since this is a bit one it could probably be split into at least two PR's:
- the hashing functions unification
- the rest

pulumi/package.json
```diff
@@ -10,10 +10,16 @@
   },
   "dependencies": {
     "@shapeshiftoss/cluster-launcher": "^0.19.0",
+    "axios": "^0.27.2",
+    "dotenv": "^16.0.3",
+    "folder-hash": "^4.0.2",
+    "object-hash": "^3.0.0",
```
kaladinlight, line 16: these exist in top level `package.json` and are accessible as hoisted node modules

pulumi/src/api.ts
```diff
@@ -18,33 +19,47 @@ export interface ApiConfig {
...
   provider: k8s.Provider
-  secretEnvs: (coinstack: string, asset: string) => k8s.types.input.core.v1.EnvVar[]
+}
+
+const getbuildAndPushImageArgs = (coinstackType: CoinstackType) => {
+  switch (coinstackType) {
+    case CoinstackType.GO:
+      return { context: '../../../', dockerFile: '../../../build/Dockerfile' }
+    case CoinstackType.NODE:
+      return { context: `../api` }
```
kaladinlight, line 38: Let's add default throw case just to cover it

pulumi/src/api.ts
```diff
@@ -18,33 +19,47 @@ export interface ApiConfig {
...
+  }
+}
+
+const getContainer = (coinstackType: CoinstackType, coinstack: string) => {
+  switch (coinstackType) {
+    case CoinstackType.GO:
+      return { container: { args: ['-swagger', 'swagger.json'] } }
+    case CoinstackType.NODE:
+      return { command: ['node', `dist/${coinstack}/api/src/app.js`] }
```
kaladinlight, line 47: default case

pulumi/src/coinstack.ts
```diff
@@ -0,0 +1,145 @@
...
+  })
+
+  const coinServices = aggregateCoinServiceInput(config.statefulService?.services || [], serviceInput)
+    .map((cs) => enrichWithPulumiConfig(cs, namespace, coinstack, config))
+    .map((cs) => createService(cs, assetName, snapshots))
+  await deployStatefulService(appName, assetName, provider, namespace, config, coinServices, volumes)
+  return {}
+}
+
+const enrichWithPulumiConfig = (
```
kaladinlight, line 46: We should rethink this. Why not include this as part of the top level service config?

kaladinlight: along these lines and can include `env` at that point
image

pulumi/src/coinstack.ts
```diff
@@ -0,0 +1,145 @@
...
+      ...csi.env,
+      L1_RPC_ENDPOINT: `http://ethereum-svc.${namespace}.svc.cluster.local:8332`,
+    }
+  }
+
+  return csi
+}
+
+// Join the config set in the Pulumi config with the typescript input from the coinstack
+const aggregateCoinServiceInput = (
```
kaladinlight, line 115: wont need this by lifting config either

pulumi/src/hash.ts
```diff
@@ -0,0 +1,133 @@
+import { createHash } from 'crypto'
+import { parse } from 'dotenv'
+import { hashElement } from 'folder-hash'
+import objectHash from 'object-hash'
+import * as k8s from '@pulumi/kubernetes'
+
+export enum CoinstackType {
```
kaladinlight, line 7: i much prefer string unions over enums: `export type CoinstackType = 'go' | 'node'`

pulumi/src/hash.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+// creates a hash of the content included in the final build image
+const getNodeCoinstackHash = async (coinstack: string, buildArgs: Record<string, string>): Promise<string> => {
+  const hash = createHash('sha1')
+  const nodeBasePath = `${__dirname}/../../node`
+
+  // hash root level unchained files
+  const { hash: unchainedHash } = await hashElement(nodeBasePath, {
```
kaladinlight, line 37: this needs to be the root dir and also include the `yarn.lock` and `Dockerfile.node`

pulumi/src/hash.ts
```diff
@@ -0,0 +1,133 @@
...
+  switch (coinstackType) {
+    case CoinstackType.NODE:
+      return await getNodeCoinstackHash(coinstack, buildArgs)
+    case CoinstackType.GO:
+      return await getGoCoinstackHash(coinstack, buildArgs)
+  }
+}
+
+// creates a hash of the content included in the final build image
+const getNodeCoinstackHash = async (coinstack: string, buildArgs: Record<string, string>): Promise<string> => {
```
kaladinlight, line 32: prefer `getNodeCoinstackApiHash` or some inclusion that this is for the api docker image hash

pulumi/src/hash.ts
```diff
@@ -0,0 +1,133 @@
...
+  })
+  hash.update(apiHash)
+
+  hash.update(objectHash(buildArgs))
+
+  return hash.digest('hex')
+}
+
+// creates a hash of the content included in the final build image
+const getGoCoinstackHash = async (coinstack: string, buildArgs: Record<string, string>): Promise<string> => {
```
kaladinlight, line 70: same naming comment

pulumi/src/index.ts
```diff
@@ -58,11 +66,34 @@ export interface ServiceConfig {
   storageSize: string
 }
 
+export interface ServiceInput {
```
kaladinlight, line 69: prefer `ServiceArgs` or can even use your coin service vernacular `CoinServiceConfig`

pulumi/src/index.ts
```diff
@@ -58,11 +66,34 @@ export interface ServiceConfig {
...
   >
   configMapData: Record<string, string>
   containers: Array<k8s.types.input.core.v1.Container>
   volumeClaimTemplates: Array<k8s.types.input.core.v1.PersistentVolumeClaim>
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Outputs = Record<string, any>
+
+export type SecretData = pulumi.Input<{
```
kaladinlight, line 97: Don't really need this typing. The secret data func can just return `Record<string, string>` (pulumi allows regular types to be used for Input<> vars)

pulumi/src/statefulService.ts
```diff
@@ -1,102 +1,81 @@
...
-  args?: Array<string>
-  env?: Record<string, string>
-  dataDir?: string
-  configMapData?: Record<string, string>
-  volumeMounts?: Array<k8s.types.input.core.v1.VolumeMount>
-  readinessProbe?: k8s.types.input.core.v1.Probe
-  livenessProbe?: k8s.types.input.core.v1.Probe
-  snapshots: VolumeSnapshot[]
-}
+export function createService(cs: JointCoinServiceInput, assetName: string, snapshots: VolumeSnapshot[]): Service {
```
kaladinlight, line 7: prefer `service` or `config` over `cs` here

pulumi/src/index.ts
```diff
@@ -58,11 +66,34 @@ export interface ServiceConfig {
...
+  dataDir?: string
+  configMapData?: Record<string, string>
+  volumeMounts?: Array<k8s.types.input.core.v1.VolumeMount>
+  readinessProbe?: k8s.types.input.core.v1.Probe
+  livenessProbe?: k8s.types.input.core.v1.Probe
+}
+
+export interface JointCoinServiceInput extends ServiceInput, ServiceConfig {}
+
 export interface Service {
```
kaladinlight, line 84: Let's move out any associated types to their corresponding files to encapsulate usage

pulumi/src/statefulService.ts
```diff
@@ -335,9 +315,9 @@ export async function deployStatefulService(
 
     const additionalRootDomainName = process.env.ADDITIONAL_ROOT_DOMAIN_NAME
 
-    const match = (service: string, prefix?: string) => {
+    const match = (service: Service, prefix?: string) => {
```
kaladinlight, line 318: did you want to clean up the domain name stuff while you're at it, or follow up pr?

pulumi/src/statefulService.ts
```diff
@@ -379,9 +359,9 @@ export async function deployStatefulService(
         },
         spec: {
           entryPoints: ['web', 'websecure'],
-          routes: Object.entries(services)
-            .map(([service, { ports }]) =>
-              ports
+          routes: services
+            .map((service) =>
```
kaladinlight, line 363: think this got broken, `service` is now the object, not the name which is used when passing to `match`

kaladinlight:
Looks solid overall. I think with lifting the config so we can have a fully explicit service config at the pulumi index.ts level, it will be :100:

pulumi/src/statefulService.ts
lmyslinski: preferably the first option

pulumi/src/coinstack.ts
lmyslinski: But that's the essence of the problem. The config parsing does not happen where input arguments are defined, so we can't reference config variables until already inside `deployCoinstack`

pulumi/src/index.ts
lmyslinski: You shouldn't really need to see what a type precisely consists of to understand a piece of code. Simultaneously, that's a good test of the abstractions we're using. If we need to inspect the type to make sense of it, that means it's not a very well-named and defined type

pulumi/src/coinstack.ts
kaladinlight: That's why I'm suggesting parsing the config one level up and passing it down to `deployCoinstack`. Then we have the `namespace` where needed and can have the env vars set explicitly for each coinstack

pulumi/src/index.ts
kaladinlight: Yeah, this is certainly dependent on the context. We have many `types.ts` files throughout the code base for shared types. Class args are a good example of something that makes sense to have alongside the class, or function param types as you stated. Service is not used outside of the `statefulService` file/types which is why I commented on this one specifically. Subjective for sure, but I find it easier to read/reference when there is some aspect of scope related to where the type is defined/exported rather than some file that is just a dumping ground for any and every type

pulumi/src/api.ts
```diff
@@ -18,33 +19,51 @@ export interface ApiConfig {
...
+      return { context: `../api` }
+    default:
+      throw new Error('invalid coinstack type')
+  }
+}
+
+const getContainer = (coinstackType: CoinstackType, coinstack: string) => {
+  switch (coinstackType) {
+    case 'go':
+      return { container: { args: ['-swagger', 'swagger.json'] } }
```
kaladinlight, line 47:
```suggestion
      return { args: ['-swagger', 'swagger.json'] }
```

lmyslinski:
> Seeing deleted swagger.json files that need to be tracked/added back

Saw that too, was looking for that thx

pulumi/src/statefulService.ts
```diff
@@ -379,13 +359,13 @@ export async function deployStatefulService(
         },
         spec: {
           entryPoints: ['web', 'websecure'],
-          routes: Object.entries(services)
-            .map(([service, { ports }]) =>
-              ports
+          routes: services
+            .map((serviceEntry) =>
```
kaladinlight, line 363: `serviceEntry` -> `service`

pulumi/src/statefulService.ts
```diff
@@ -402,7 +382,7 @@ export async function deployStatefulService(
             .flat(),
           tls: {
             secretName: secretName,
-            domains: Object.keys(services).map((service) => ({ main: domain(service) })),
+            domains: services.map((serviceEntry) => ({ main: domain(serviceEntry) })),
```
kaladinlight, line 385: same naming comment

pulumi/src/statefulService.ts
kaladinlight, line 318: if you are going to pass the full service, you need to use `service.name` for all the string interpolation. Could just leave it as a string and pass in `service.name` in the map though makes a bit more sense to me

pulumi/src/statefulService.ts
lmyslinski: Did that on purpose to avoid variable shadowing. `service` is already defined in this scope

pulumi/src/statefulService.ts
lmyslinski: > you need to use service.name
Isn't that the case right now?

lmyslinski: The latter decreases type safety, which in case of this specific function I'd rather avoid. I have a hard time navigating this 200 line context as is

pulumi/src/index.ts
```diff
@@ -58,11 +66,34 @@ export interface ServiceConfig {
...
+  args?: Array<string>
+  env?: Record<string, string>
+  dataDir?: string
+  configMapData?: Record<string, string>
+  volumeMounts?: Array<k8s.types.input.core.v1.VolumeMount>
+  readinessProbe?: k8s.types.input.core.v1.Probe
+  livenessProbe?: k8s.types.input.core.v1.Probe
+}
+
+export interface JointCoinServiceInput extends ServiceInput, ServiceConfig {}
```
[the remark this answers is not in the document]
lmyslinski: Not quite, they both have required props so we can't loosely swap them

pulumi/src/coinstack.ts
lmyslinski: I did the lift but I think this needs to stay. Can't have a unified type for both of these, both have unique required args

pulumi/src/index.ts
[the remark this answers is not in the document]
kaladinlight: `interface JointCoinServiceInput extends ServiceInput, ServiceConfig {}` is the same as
```
interface ServiceInput extends ServiceConfig {
   ...serviceInputFields
}
```

pulumi/src/statefulService.ts
kaladinlight: then we need `service.name` for the string interpolation. will make another comment

pulumi/src/statefulService.ts
kaladinlight: that's fine, not super concerned about he shadowing (I feel like there is a lot of that in js specifically with these kind of callback functions. could use `svc` or something, but your call. just feel like stylistically it is usual plural.map((singular/shortened))

pulumi/src/index.ts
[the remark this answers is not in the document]
lmyslinski: Then how are you going to compile the `ServiceInput`/`ServiceArgs`? 

```
Type '{ coinServiceName: string; ports: { 'daemon-rpc': { port: number; }; }; env: { NETWORK: string; }; }' is missing the following properties from type 'ServiceArgs': image, name, storageSizets(2739)
```
Please push this change if you see it can work, I still don't

kaladinlight (approved):

PR #690, opened 2023-04-14
fix: eth cron

lmyslinski:
Set the correct cron format


gomesalexandre (approved):
4AM on Sunday it is then

PR #698, opened 2023-04-20
feat: biweekly backups for the bad kids

lmyslinski:
Because both polygon and bnb are more likely to fail than other chains, it would be wise to snapshot them more often and keep 2 backups instead of one. Sat and wed seem like good dates, as we're unlikely to not notice broken chains over a week's period


gomesalexandre (approved):
04:00 AM, only on Wednesday and Saturday it is then

kaladinlight (approved):
seems like a good starting point to me!

PR #704, opened 2023-04-25
fix: added error snapper handling and fixed namespace issue

lmyslinski:
- Add namespace arg to make sure volume snapshots are made in the correct namespace
- Added error handling so that even when a snapshot fails, the stateful set is scaled back up

### DIFF
--- a/pulumi/src/snapper.ts
+++ b/pulumi/src/snapper.ts
@@ -74,9 +74,9 @@ export class Snapper {
 
   protected async takeSnapshots(pvcList: Array<string>): Promise<void> {
     const timestamp = new Date()
-    await Promise.allSettled(
-      pvcList.map(async (pvc) => {
-        try {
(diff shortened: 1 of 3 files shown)


pulumi/src/snapper.ts
```diff
@@ -72,8 +73,16 @@ export class Snapper {
   }
 
   protected async takeSnapshots(pvcList: Array<string>): Promise<void> {
-    const timestamp = new Date()
+    try {
+      return await this.takeSnapshotInternal(pvcList)
+    } catch (err) {
+      console.error(`Could not create VolumeSnaphot:`, err)
+      return Promise.resolve()
```
kaladinlight, line 80: not required if you want to omit

kaladinlight (approved):

kaladinlight (approved):

PR #714, opened 2023-04-27
chore: preview coinstack changes to dev env on feature branches

lmyslinski:
We currently only lint the config on our PR's - we could also add running `pulumi preview` against `dev` environment so that we can see what would the impact be after merging said changes. This should shorten our dev loop, reduce the amount of errors we introduce and doesn't add any additional cost.


lmyslinski:
Figured out a simpler solution

PR #721, opened 2023-05-01
feat: gas oracle

kaladinlight:
- add gas price oracle to cache blockchain state for estimations
- transition from `eth_feeHistory` to gas price oracle that can work for both eip1559 and legacy fees
- add legacy gas price estimation
- keep 20 blocks and estimate using 20 block average (this can be extended if required, but suits our needs for now)
- includes state validation to ensure we always have 20 blocks and they are the most recent 20 blocks including pending
- updates over websocket are primary intent, but validation will ensure accurate data on request as well for a fallback


node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    const estimatedFees = percentiles.reduce<EstimatedFees>((prev, percentile) => {
+      const { gasPrice, maxPriorityFee } = this.averageAtPercentile(percentile)
+      prev[percentile.toString()] = {
+        gasPrice: gasPrice.toFixed(0),
+        maxFeePerGas: (maxPriorityFee + Number(baseFeePerGas)).toFixed(0),
+        maxPriorityFeePerGas: maxPriorityFee.toFixed(0),
+      }
+      return prev
+    }, {})
```
lmyslinski, line 75: Nitpick: can we use a more meaningful name that `prev` in `reduce` blocks? Reduce always fucks with my brain a bit, something like `estimations[percentile.toString()]` is a lot more meaningful out of the box.

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,162 @@
...
+  // estimate fees based on the current oracle state
+  async estimateFees(percentiles: Array<number>): Promise<EstimatedFees> {
+    const { baseFeePerGas, number } = (await this.provider.send('eth_getBlockByNumber', ['pending', false])) as {
+      number: string
+      baseFeePerGas?: string
+    }
+
+    const latestBlock = Number(number) - 1
+
+    await this.validate(latestBlock)
```
lmyslinski, line 56: Running the validation with the update logic on estimation can potentially take a very long time. Probably better to check if all data up to date, if yes then pass validation and get estimate, else return fallback and update in the bg?

lmyslinski (changes requested):
I like the way the flow is organized but I think we can give it a bit more polish in terms of code organization:

- `gasOracle.ts` should be split into two bits, one consisting of the top-level fee retrieval/estimation logic, the other of the fee-block state manipulation and management
- This type of math arithmetic with lots of `fori` loops, `+1, -1` etc. is almost impossible to get right without unit testing. I think in this case writing a few units just to make sure we're calculating the blocks correctly, not running the index out of bounds etc. is totally worth it

node/coinstacks/common/api/src/evm/gasOracle.ts
kaladinlight: The thought here was during normal runtime, the websocket updates should keep state up to date and `validate` will only take a handful of ms to run. Worst case, this would perform the duties of the websocket if we lost connection or something and fetch the necessary data from the node similarly to how we were previously getting gas prices from the node.

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,162 @@
...
+    this.logger = args.logger.child({ namespace: ['gasOracle'] })
+    this.provider = args.provider
+  }
+
+  async start() {
+    const height = await this.provider.getBlockNumber()
+    const length = this.totalBlocks - 1
+    const startingBlock = height - length
+    const blocks = Array.from<number, string | number>({ length }, (_, index) => index + 1 + startingBlock)
+    blocks.push('pending')
```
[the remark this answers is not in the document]
kaladinlight: this has been simplified a lot and shouldn't be as hard to digest now. not sure it's worth the abstraction imo

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
...
+  private baseFeePerGas?: string
+
+  constructor(args: GasOracleArgs) {
+    this.logger = args.logger.child({ namespace: ['gasOracle'] })
+    this.provider = args.provider
+    this.coinstack = args.coinstack
+    this.totalBlocks = args.totalBlocks ?? 20
+
+    // specify latestBlockTag as supported by the coinstack node
+    switch (args.coinstack) {
```
kaladinlight, line 48: exhaustive switch to force explicit coinstack check

node/packages/blockbook/src/websocket.ts
```diff
@@ -114,7 +114,12 @@ export class WebsocketClient {
       if (!res.data) return
 
       if (res.id === 'newBlock' && 'hash' in res.data) {
-        await this.handleBlock(res.data)
+        const newBlock = res.data
+        if (Array.isArray(this.handleBlock)) {
+          await Promise.all(this.handleBlock.map(async (handleBlock) => handleBlock(newBlock)))
```
kaladinlight, line 119: allSettled

node/packages/blockbook/src/websocket.ts
```diff
@@ -114,7 +114,12 @@ export class WebsocketClient {
       if (!res.data) return
 
       if (res.id === 'newBlock' && 'hash' in res.data) {
```
kaladinlight, line 116: switch

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
kaladinlight, line 1:
```suggestion
```

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
...
+      case 'optimism':
+        this.latestBlockTag = 'latest'
+        break
+      default:
+        this.latestBlockTag = 'pending'
+    }
+  }
+
+  async start() {
+    const height = await this.provider.getBlockNumber()
```
kaladinlight, line 59: validate pending block vs block number

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
...
+    const length = this.latestBlockTag === 'pending' ? this.totalBlocks - 1 : this.totalBlocks
+    const startingBlock = height - length
+    const blocks = Array.from<number, BlockTag | number>({ length }, (_, index) => index + startingBlock + 1)
```
kaladinlight, line 62: one more comment for posterity

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
...
+    }
+  }
+
+  // update oracle state for the specified block
+  private async update(blockNumber: BlockTag | number) {
+    try {
+      const numOrTag =
+        blockNumber === this.latestBlockTag
+          ? blockNumber
+          : ethers.utils.hexStripZeros(ethers.utils.hexlify(blockNumber))
```
kaladinlight, line 142: hexValue

node/coinstacks/common/api/src/evm/gasOracle.ts
```diff
@@ -0,0 +1,245 @@
...
+  private averageAtPercentile(blockFees: Array<BlockFees>, percentile: number) {
+    const valueAtPercentile = (fees: Array<number>) => {
+      if (!fees.length) return 0
+      const rank = Math.ceil((percentile / 100) * fees.length)
+      const value = fees[rank - 1]
+      return value
+    }
+
+    const sum = { gasPrice: 0, maxPriorityFee: 0 }
+    blockFees.forEach((fees) => {
```
kaladinlight, line 233: reducey for funzies

PR #724, opened 2023-05-05
chore: add nft support

kaladinlight:
- update blockbook client types
- use new values from blockbook to return nft ids and values from unchained api


node/coinstacks/common/api/src/evm/service.ts
```diff
@@ -357,16 +357,45 @@ export class Service implements Omit<BaseAPI, 'getInfo'>, API {
...
-      })),
+      tokenTransfers: tx.tokenTransfers?.map((tt) => {
+        const value = (() => {
+          switch (tt.type) {
+            case 'ERC721':
+            case 'BEP721':
+              return '1'
+            case 'ERC1155':
+            case 'BEP1155':
+              return tt.multiTokenValues?.[0].value ?? '0'
```
kaladinlight, line 368: multi tokens can be multiple of the same id

node/coinstacks/common/api/src/evm/service.ts
```diff
@@ -357,16 +357,45 @@ export class Service implements Omit<BaseAPI, 'getInfo'>, API {
...
+            default:
+              return tt.value
+          }
+        })()
+
+        const id = (() => {
+          switch (tt.type) {
+            case 'ERC721':
+            case 'BEP721':
+              return tt.value
```
kaladinlight, line 378: blockbook returns the nft id under `value`, use as `id`

node/coinstacks/common/api/src/evm/service.ts
```diff
@@ -357,16 +357,45 @@ export class Service implements Omit<BaseAPI, 'getInfo'>, API {
...
+            case 'ERC1155':
+            case 'BEP1155':
+              return tt.multiTokenValues?.[0].id
```
kaladinlight, line 381: get multi token id (not sure why it was set up as an array on the blockbook side, only ever length of 1 at this time)

node/coinstacks/common/api/src/evm/service.ts
gomesalexandre: 0 looks correct here, though do we want to guard against undefined `multiTokenValues[0]` as a paranoia check?
https://github.com/trezor/trezor-suite/blob/ddd24ef647b9d088c26c0d337cd2de3c9d4b1d56/suite-common/wallet-utils/src/transactionUtils.ts#L413-L416

node/coinstacks/common/api/src/evm/service.ts
gomesalexandre: Ditto, referring to trezor-suite as the source of truth 0 looks correct, though we might want to guard against `multiTokenValues[0]` being defined, or alternatively, `multiTokenValues?.length`

gomesalexandre (commented):
Not an unchained wizard but cross-referencing the blockbook data consumption in trezor-suite, LGTM

node/coinstacks/common/api/src/evm/service.ts
kaladinlight: they are just checking length and then grabbing id, I think we are good unless we see the api change

kaladinlight: meh, doesn't hurt, adding lol

node/coinstacks/common/api/src/evm/service.ts
kaladinlight: doin it

gomesalexandre (approved):
Conceptual stamp

PR #747, opened 2023-05-16
feat: discord alerts

lmyslinski:
This PR introduced alerts about the state of the unchained backend infrastructure. This setup is based on a modified version of [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus). 

- The `build.sh` script injects secret variables from env into the alertmanager config template, generating the final config.

Alert manager configuration details:
- kube-prometheus comes with a ton of preconfigured alerts, some of which are firing all the time. Instead of removing them, we can only select which ones we're interesed in.
- It is not possible to send an alert exactly once, however instead we set the interval to a long enough interval so that we're not spammed when things go wrong. 
- Alert manager is configured so that we're not firing the same alert multiple times with different severity (an alert might be triggered multiple times, based on thresholds ex. 15% for warning, 50% for critical)
- Non-listed alerts are silenced (sent to blackhole receiver)
- Alert manager will group several alerts of the same type into a single message

Initially only 3 alerts are allowed:
- volume filling up - in case we're running out of disk space
- job failed - in case a backup job didn't complete successfully
- cpu throttling - turned on just for prod testing, it's unlikely to be resolved on its own soon. We should probably tackle this sometime 

Sample alert:
Screenshot from 2023-05-24 15-35-42


monitoring/src/config.ts
```diff
@@ -30,6 +30,8 @@ export const getConfig = async (): Promise<LoopConfig> => {
   const domain = (await stackReference.getOutputValue('rootDomainName')) as string
   const additionalDomain = process.env.ADDITIONAL_ROOT_DOMAIN_NAME
 
+  console.log(`Extra domain: ${additionalDomain}`)
```
NeOMakinG, line 33: Gm! Is it wanted?

lmyslinski: ...not quite. I've left this here just to demonstrate an issue with the current CI workflow, where even thou our domains are seemingly injected correctly, the resulting ingress does not have that domain on the match list, resulting in a 404 once you visit https://monitoring.shapeshift.com/. Because this flow is only triggered on demand, I was trying to replicate the issue and see if there's some condition to the failures, so far with no lock. So short answer is no, but I'd keep this in until this issue is found and fixed

lmyslinski: Reverting this to make this PR posh, will work on this separately

monitoring/my-kube-prometheus/readme.md
```diff
@@ -10,6 +10,17 @@ The state is not kept in the repository, as it's a lot of generated yaml files.
 
 This setup works in tandem with the existing Pulumi config, and the Pulumi monitoring config should be applied directly after deploying the changes configured here.
 
+# Initial setup
+
+Install [goland](https://go.dev/)
+Install [jsonnet](https://github.com/google/jsonnet)
```
kaladinlight, line 16:
```suggestion
```

monitoring/my-kube-prometheus/default.tmpl
```diff
@@ -0,0 +1,18 @@
...
+{{ define "our.discord.title" }}
+    Unchained alert! {{ .GroupLabels.SortedPairs.Values | join " " }}
+{{ end }}
+
+{{ define "our.discord.message" }}
+    {{ range .Alerts.Firing }}
+      Alert: **{{ printf "%.150s" .Annotations.summary }}** ({{ .Labels.severity }})
+      Description: {{ printf "%.150s" .Annotations.description }}
+      Alertname: {{ .Labels.alertname }}
+      Namespace: {{ .Labels.namespace }}
```
kaladinlight, line 12: add service label? seeing them as empty on the alerts in the discord channel

monitoring/my-kube-prometheus/alertmanager-config.tmpl.yaml
```diff
@@ -0,0 +1,36 @@
+route:
+  group_by: ['alertname', 'severity']
+
+  group_wait: 30s
+  # Only send the same alerts every 12h
+  group_interval: 12h
+  repeat_interval: 12h
+  # We must deliver all alerts somewhere, so we use send them to a blackhole
```
kaladinlight, line 8:
```suggestion
  # We must deliver all alerts somewhere, so we send them to a blackhole
```

monitoring/my-kube-prometheus/alertmanager-config.tmpl.yaml
```diff
@@ -0,0 +1,36 @@
...
+receivers:
+- name: discord
+  discord_configs:
+  - webhook_url: <<discord_webhook_url>>
+    message: '{{ template "our.discord.message" . }}'
+    title: '{{ template "our.discord.title" . }}'
+
+- name: blackhole
+  
+
```
kaladinlight, line 34:
```suggestion
```

monitoring/my-kube-prometheus/default.tmpl
lmyslinski: Cleaned this one up already but didn't apply it

kaladinlight (approved):

PR #761, opened 2023-05-25
feat: expose prometheus

lmyslinski:
Expose our [prometheus](https://prometheus.io/) instance under a public domain `prometheus.shapeshift.com`. Prometheus allow for easy access to all of our alerts, so that we have a detailed look at what exactly is firing and where. Because prometheus does not offer any auth solutions, it is secured via basic-auth with the use of a Traeffic ingress. The credentials are injected as part of our CI pipeline.


monitoring/src/prometheus.ts
```diff
@@ -0,0 +1,135 @@
...
+      {
+        apiVersion: 'traefik.containo.us/v1alpha1',
+        kind: 'Middleware',
+        metadata: {
+          name: `${name}-prometheus-ingress-auth`,
+          namespace: args.namespace,
+        },
+        spec: {
+          basicAuth: {
+            secret: `${name}-prometheus-auth-secret`,
```
kaladinlight, line 75: you can assign the resource to a variable and access the metadata for resource names as an alternative to reduce possible name change discrepancy in the future `secret.metadata.name`. probably won't ever change but just an fyi for reference

monitoring/src/prometheus.ts
```diff
@@ -0,0 +1,135 @@
...
+        },
+        spec: {
+          entryPoints: ['web', 'websecure'],
+          routes: [
+            {
+              match: domains,
+              kind: 'Rule',
+              middlewares: [
+                {
+                  name: `${name}-prometheus-ingress-auth`,
```
kaladinlight, line 99: likewise here

monitoring/src/prometheus.ts
```diff
@@ -0,0 +1,135 @@
...
+                  namespace: args.namespace,
+                },
+              ],
+              services: [
+                {
+                  kind: 'Service',
+                  name: `prometheus-k8s`,
+                  port: 9090,
+                  namespace: `${name}-monitoring`,
```
kaladinlight, line 108:
```suggestion
                  namespace: args.namespace,
```

kaladinlight (approved):

PR #763, opened 2023-05-29
fix: use correct = operator in bash

lmyslinski:
Our readiness checks incorrectly compare strings, resulting in an error:

```
root@gnosis-sts-0:/nethermind# cat test.sh
#!/bin/sh

ETH_SYNCING=$(curl -s http://localhost:5052/eth/v1/node/syncing)
SYNCING_RESULT=$(echo "$ETH_SYNCING" | jq -r .data.is_syncing)

if [ "$SYNCING_RESULT" == "false" ]; then
  echo "node is synced"
else
  echo "node is still syncing"
  exit 1
fi

root@gnosis-sts-0:/nethermind# ./test.sh
./test.sh: 6: [: true: unexpected operator
node is still syncing
root@gnosis-sts-0:/nethermind#
```

After the patch:
```
root@gnosis-sts-0:/nethermind# cat test.sh
#!/bin/sh

ETH_SYNCING=$(curl -s http://localhost:5052/eth/v1/node/syncing)
SYNCING_RESULT=$(echo "$ETH_SYNCING" | jq -r .data.is_syncing)

if [ "$SYNCING_RESULT" = "false" ]; then
  echo "node is synced"
else
  echo "node is still syncing"
  exit 1
fi

root@gnosis-sts-0:/nethermind# ./test.sh
node is still syncing
```
This doesn't alter the execution but gets rid of the error


gomesalexandre (approved):
Will let @kaladinlight confirm but according to my bash fu (a.k.a LLM fu) LGTM

lmyslinski:
Thanks for the context, will patch as suggested and then get back to it

lmyslinski:
@kaladinlight updated

kaladinlight (approved):

PR #767, opened 2023-06-02
chore: trace method for internal tx

lmyslinski:
Introduces an additional RPC method for fetching internal tx's. By default all existing transactions use the `debug` mode, while for `gnosis` we use the new `trace` mode to address current performance problems.

The trace module has been confirm to be much less resource heavy. The memory usage never spiked over 1 MB after the change. 

Local test result with Gnosis:
Zrzut ekranu 2023-06-2 o 18 29 22


kaladinlight (approved):

lmyslinski:
Also verified existing debug module still works as expected:
image

PR #816, opened 2023-07-23
chore: capture current restore state config used

0xdef1cafe:
do not merge - for posterity only


kaladinlight:
https://github.com/shapeshift/unchained/pull/824 captures an example yaml

PR #876, opened 2023-10-25
chore: infra maintenance

kaladinlight:
- right size and upgrade services
- add env var for polkachu api upgrade
- update geth to pebbledb
- add missing aria2 dependency for polygon snapshots
- update avalanche to archive node config


.circleci/config.yml
```diff
@@ -1361,56 +1361,56 @@ workflows:
...
-
-      - deploy-coinstack-node:
-          name: deploy ethereum
-          organization: TAXISTAKE
-          pulumi-command: up -f --yes
-          requires:
-            - approve ethereum coinstack
-          <<: *ethereum
-          <<: *only-main
+      #- deploy-coinstack-node:
```
0xApotheosis, line 1364: I don't have the context here, but checking this is meant to be commented out?

kaladinlight: Good eyes and question. So I am in the middle of trying to upgrade ethereum to pebbledb and avalanche to a full archive node which requires volume restores and daemon config changes for both. I have only gotten develop rolling and wanted to make sure we couldn't accidentally deploy the changes to prod before it was fully set up. This will be reverted once complete and deploys are safe and happy again

PR #928, opened 2024-01-24
fix: resolves issues when running in docker on mac

0xean:
Attempting to run lerna watch locally in docker would fail due to how we copied over node_modules into the linux container if using a mac.  Adding yarn install to the compose enforces that the platform specific version of @nx/nx-linux-arm64-musl is installed and seems to fix the issues.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,10 @@ services:
       context: ./node
       dockerfile: Dockerfile.local
     working_dir: /app/node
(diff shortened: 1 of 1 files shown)


kaladinlight (approved):
Thanks @0xean! This should not add too much overhead for now and we can improve things further when we cut unchained over to use yarn 3 and leverage yarn workspaces over lerna.

PR #1047, opened 2024-09-24
feat: add zrx proxy

kaladinlight:
- Supports both v1 and v2 apis.
  - v1 leverages chain path for correct routing
  - v2 leverages `chainId` query param for correct routing
- No backend response cache at this time


kaladinlight:
closes: https://github.com/shapeshift/web/issues/7784

node/proxy/api/src/zrx.ts
```diff
@@ -0,0 +1,80 @@
...
+  constructor() {
+    this.axiosInstance = axios.create({
+      headers: {
+        '0x-api-key': ZRX_API_KEY,
+      },
+    })
+  }
+
+  async handler(req: Request, res: Response): Promise<void> {
+    const parsedUrl = new URL('https://dummy.com'.concat(req.url))
```
gomesalexandre, line 21: Is there any rationale here re: not using req.url directly?

node/proxy/api/src/zrx.ts
```diff
@@ -0,0 +1,80 @@
...
+    const path = parsedUrl.pathname.substring('/api/v1/zrx/'.length)
```
gomesalexandre, line 22: 🧠

gomesalexandre: Could potentially use a capture group to make this less leetcody magic but meh

node/proxy/api/src/zrx.ts
kaladinlight: `req.url` doesn't include the base url and will not be parsed correctly by `URL`, so just added a dummy base url so we can use it and make sure there is no funny business with pulling off the query params (versus just split on `?`)

node/proxy/api/src/zrx.ts
kaladinlight: was following the pattern, though not sure how I ended up with substring at one point lol. I will just update with a simple `.replace()`

PR #1049, opened 2024-09-26
feat: add solana send endpoint

### LINKED ISSUE #1031: solana send endpoint
- send endpoint
- copy existing controller w/ send route 
- nodejs impelementation (is there an existing SDK  or should we just use axios)

NeOMakinG:
closes #1031

## Description

Added the send endpoint logic:
- I decided to use the helius SDK because I feel like their custom endpoints will be typed this way, it will be easier for us to respect their types
- Like other chains, I assumed the TX is signed on client side

## Testing
- You need to sign a TX
``` ts
import { Transaction, SystemProgram, Connection, PublicKey } from '@solana/web3.js';

 ;(async () => const transaction = new Transaction().add(
    SystemProgram.transfer({
      fromPubkey: publicKey,
      toPubkey: publicKey,
      lamports: 100,
    })
  );
  transaction.feePayer = publicKey;

  const anyTransaction: any = transaction;
  anyTransaction.recentBlockhash = (await connection.getLatestBlockhash()).blockhash;
  
   try {
    const signedTransaction = await provider.signTransaction(transaction);
    console.log(Buffer.from(signedTransaction.serialize()).toString('base64'))
  } catch (error) {
    console.warn(error);
    throw new Error(error.message);
  }
)()
```

Then use this with the send endpoint (http://api.solana.localhost/api/v1/send)

## Screenshots
image

Here is the TX:
https://solscan.io/tx/2yG59F1eughDiEg6R8FkcDe5MVSd2YYqiBeaH1AMX958nMXrLEpVPvgKpCDFwY2Doh6oM7NeHNdSbrcfSbckv8Ps


kaladinlight:
A couple comments:
- Just to give context for the `Service` paradigm. The `tsoa` route registration does not support constructor ags. The service was a way to abstract shared logic between evm/utxo chains and pass args to the class for the specific endpoints/api keys. Seeing as this isn't really the case with solana and we could probably reduce the extra boilerplate and just implement the functions in `controller.ts` directly.
- We can use the helius sdk for now and agree it is necessary for typed responses from their api. Note that the `connection` is just an instance of `@solana/web3.js` connection and we _could_ use web3 directly for rpc/subscription related things. May not be worth is the heluis sdk is more ergonomic to use though.

node/coinstacks/solana/api/src/controller.ts
```diff
@@ -78,8 +91,9 @@ export class Solana implements BaseAPI {
...
   @Response<InternalServerError>(500, 'Internal Server Error')
   @Post('send/')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async sendTx(@Body() _body: SendTxBody): Promise<string> {
-    return ''
+    return Solana.service.sendTx(_body)
   }
 }
+
+Solana.service = service
```
kaladinlight, line 99: vote to get rid of service to reduce boilerplate as I described in the comments

node/coinstacks/solana/api/src/service.ts
```diff
@@ -0,0 +1,38 @@
...
+export interface ServiceArgs {
+  rpcUrl: string
+  rpcApiKey: string
+}
+
+export class Service implements Omit<BaseAPI, 'getInfo'>, API {
+  private readonly heliusSdk: Helius
+
+  constructor(args: ServiceArgs) {
+    this.heliusSdk = new Helius(args.rpcApiKey)
```
kaladinlight, line 14: we should pass rpc url in case we ever point to nownodes or something else

node/coinstacks/solana/api/src/service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  async getAccount(): Promise<Account> {
+    throw new Error('Method not implemented.')
+  }
+
+  async getTxHistory(): Promise<TxHistory> {
+    throw new Error('Method not implemented.')
+  }
+
+  async sendTx(body: SendTxBody): Promise<string> {
```
kaladinlight, line 29: i wish I would have use `body.rawTx` or something more generic since `hex` doesn't really make sense for solana... can make a breaking change to update or just leave as is for now.

node/coinstacks/solana/api/src/service.ts
NeOMakinG: Yeah this is kinda why I kept them in the args, but I couldn't instantiate them without adding an eslint disable comment, anyway it will be removed by the service removal naturally 👍

node/coinstacks/solana/api/src/controller.ts
```diff
@@ -78,8 +91,9 @@ export class Solana implements BaseAPI {
   @Response<ValidationError>(422, 'Validation Error')
   @Response<InternalServerError>(500, 'Internal Server Error')
   @Post('send/')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async sendTx(@Body() _body: SendTxBody): Promise<string> {
```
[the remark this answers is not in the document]
NeOMakinG: Done in [412c348](https://github.com/shapeshift/unchained/pull/1049/commits/412c348ee52620a3028caadd8322dd58348bbd20)

node/coinstacks/solana/api/src/controller.ts
NeOMakinG: Done in [412c348](https://github.com/shapeshift/unchained/pull/1049/commits/412c348ee52620a3028caadd8322dd58348bbd20)

node/coinstacks/solana/api/src/service.ts
NeOMakinG: Done in [412c348](https://github.com/shapeshift/unchained/pull/1049/commits/412c348ee52620a3028caadd8322dd58348bbd20)

node/coinstacks/solana/api/src/service.ts
NeOMakinG: For the sake of simplicity I would keep it as it is, we don't necessary need to change it now but in a followup or another ticket related to this

I would prefer to have the diff for the send endpoint in this commit and another commit for the refactoring so in case any thing occur it's simpler to revert or investigate

NeOMakinG:
> A couple comments:
> 
> * Just to give context for the `Service` paradigm. The `tsoa` route registration does not support constructor ags. The service was a way to abstract shared logic between evm/utxo chains and pass args to the class for the specific endpoints/api keys. Seeing as this isn't really the case with solana and we could probably reduce the extra boilerplate and just implement the functions in `controller.ts` directly.
> * We can use the helius sdk for now and agree it is necessary for typed responses from their api. Note that the `connection` is just an instance of `@solana/web3.js` connection and we _could_ use web3 directly for rpc/subscription related things. May not be worth is the heluis sdk is more ergonomic to use though.

The helius sdk will be more ergonomic to use because of the in houses endpoints that are not usable with web3js

node/coinstacks/solana/api/src/service.ts
NeOMakinG: I wonder if we really needed this? 

I would completely separate NOWNodes instances from the Helius SDK and use web3js with NOWNodes

The Helius SDK is adding another parameter (the Helius API key) to the RPC URL which we don't need with NOWNodes + it adds in houses methods that we don't need with a regular RPC

node/coinstacks/solana/api/src/service.ts
kaladinlight: Agreed, sounds like a plan

node/coinstacks/solana/api/src/service.ts
kaladinlight: Yeah I guess that was the point I was trying to get at. To me it feels like we should use web3 for raw jsonrpc calls so that we may switch out rpc urls/api keys as needed, and helius sdk for actual helius api calls. But for now, happy to send things as is, not like it would be a hard update if/when we ever need to use nownodes + helius

kaladinlight (approved):

PR #1055, opened 2024-10-08
feat: solana getAccount

kaladinlight:


node/coinstacks/solana/api/src/controller.ts
```diff
@@ -64,7 +65,48 @@ export class Solana implements BaseAPI, API {
...
+      let balance = '0'
+      let hasMore = false
+      let tokens: Array<TokenBalance> = []
+
+      do {
+        const { total, limit, items, nativeBalance } = (await heliusSdk.rpc.getAssetsByOwner({
+          ownerAddress: pubkey,
+          page: page++,
+          displayOptions: { showFungible: true, showNativeBalance: true, showZeroBalance: true },
+        })) as DAS.GetAssetResponseList & { nativeBalance: NativeBalance }
```
NeOMakinG, line 79: god... :(

NeOMakinG (approved):
## small account
image

## VERY big account

I did test with `BobogahEps526rjHMsYFoqXD8d7dzFGcmB9M5LXj5QX1` which is a famous bot on pumpfun, it took something like 40sec to load his account but it worked, swagger doc couldn't display the result thought but the request was succesful

Looks good to me!

node/coinstacks/solana/api/src/controller.ts
kaladinlight: yeah, they added the `displayOption` key and then didn't update the response payload... definitely a bit of amateur hour over there lol

PR #1079, opened 2024-12-06
chore: fix typo

cuiyourong:


gomesalexandre:
https://github.com/shapeshift/unchained/blob/develop/CONTRIBUTING.MD#warning-typo-only-changes-not-accepted

plajjan:
@gomesalexandre hi, why do you not accept typo only changes? The same user opened a PR on one of my repos with a typos-only fix and I'm trying to understand this behavior. Seems bot-like but I don't quite get why? Is this why you have such policy in place? thanks!

PR #1278, opened 2026-06-12
fix(evm): validate NFT media/metadata URLs before fetching

0xApotheosis:
## Description

`GET /api/v1/metadata/token` infers a media type by fetching a URL taken straight from NFT metadata (`image`/`animation_url`, and the on-chain `tokenURI()` in the blockbook path). Those URLs are unvalidated and redirects are followed by default, so the server can end up making requests to unintended hosts.

- Add `assertSafeOutboundUrl` in `common/api/src/utils.ts`: allow only `http(s)`, resolve the host, and reject private/loopback/link-local/CGNAT and metadata-service ranges.
- Call it before each fetch and set `maxRedirects: 0`.

## Notes

- `maxRedirects: 0` means a media URL that legitimately redirects now resolves to no `media.type` instead of following the hop. That field is already best-effort (errors are swallowed to `undefined`), so this is a minor, safe degradation.

## Verification

Didn't run build/tests locally per the repo CLAUDE.md. It's a pre-fetch URL check plus disabling redirects; no change to well-formed public media URLs that respond directly.


0xApotheosis:
Superseded by #1280 (branch renamed).

PR #1290, opened 2026-09-08
fix(thorchain): recover indexed transactions missing from direct lookup

0xApotheosis:
THORChain can return a transaction in account history while `/api/v1/tx/{txid}` fails with HTTP 500: the upstream `/tx` lookup says it is missing, but `/tx_search` on the same provider still returns the complete signed transaction. For example, this affects `D4C2B478C69D020386D027CB2205CFCAE60D1BD791E93E31ACDAE8ACCB32AA9F`.

Fall back to a one-result exact-hash search only for the specific transaction-not-found RPC error and a valid 32-byte hash. Accept the result only when both its reported hash and the hash of its signed bytes match the requested transaction. Other upstream failures retain their existing behavior. Also preserve the decoding error instead of dereferencing a nil RPC error on malformed success responses.

Validation: live read-only checks recovered both audited missing transactions through exact-hash search, verified their signed-byte hashes, fetched their blocks, and confirmed account history formats the same transactions. An unknown hash returned no search results. `git diff --check` passes. Local tests, Go builds/vet, and lint were not run per `CLAUDE.md`; CI provides build/lint validation.

This addresses the THORChain lookup failure. The separately observed Arbitrum/Moralis miss is not changed here.

### DIFF
--- a/go/pkg/thorchain/tx.go
+++ b/go/pkg/thorchain/tx.go
@@ -1,7 +1,9 @@
 package thorchain
 
 import (
+	"bytes"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


kaladinlight:
Upstream bug which has been fixed.

